PR #3, opened 2022-10-25
Add GoLow method to high-level SchemaProxy

travisnewhouse:
A client of high-level SchemaProxy may want to know whether a schema is a reference and the name of the reference.  This information is available from methods GetSchemaReference() and IsSchemaReference() in low-level SchemaProxy.  Provide a GoLow() method to retrieve the low-level SchemaProxy for a high-level SchemaProxy.

### DIFF
--- a/datamodel/high/base/schema_proxy.go
+++ b/datamodel/high/base/schema_proxy.go
@@ -67,3 +67,10 @@ func (sp *SchemaProxy) Schema() *Schema {
 func (sp *SchemaProxy) GetBuildError() error {
 	return sp.buildError
 }
+
+func (sp *SchemaProxy) GoLow() *base.SchemaProxy {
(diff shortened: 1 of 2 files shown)


travisnewhouse:
@daveshanley I found this change useful to be able to get the name of a schema reference from the high-level model.  An alternative could have been to add GetSchemaReference() and IsSchemaReference() methods to high-level SchemaProxy.  Please let me know which design you think is best.

daveshanley:
Oh wow! sorry I didn't see this PR, I already made the change for you - not knowing this was already sitting here.

daveshanley (approved):
Looks good, thank you for your contribution!

daveshanley:
I have removed my quick change in favor of this tested code. Thank you for making libopenapi better!

PR #7, opened 2022-11-04
feat: add extensions to info

danielgtaylor:
From the [Open API 3.1 spec](https://spec.openapis.org/oas/latest.html#info-object) on the top-level `info` object:

> This object MAY be extended with [Specification Extensions](https://spec.openapis.org/oas/latest.html#specificationExtensions).

This PR implements support for that, which is relevant for supporting [Restish Open API extensions](https://rest.sh/#/openapi?id=name) which let you e.g. modify the name, description, etc or hide operations from the CLI directly from your Open API spec.

I added a basic test which covers the newly added lines. I see the Go formatter changed a couple of things - let me know if that's a problem and I'll undo those changes.

### DIFF
--- a/datamodel/high/base/info.go
+++ b/datamodel/high/base/info.go
@@ -4,6 +4,7 @@
 package base
 
 import (
+	"github.com/pb33f/libopenapi/datamodel/high"
 	low "github.com/pb33f/libopenapi/datamodel/low/base"
 )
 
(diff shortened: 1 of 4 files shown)


daveshanley (approved):
Looks good! I knew I had missed a couple of these extension points! Thank you for making libopenapi better and for your contribution!

PR #16, opened 2022-11-28
fix: conditionally set min/max/multiple

danielgtaylor:
It looks like the high level schema `Minimum`, `Maximum`, and `MulitpleOf` would always get set to `int64(0)` which, unfortunately, is a valid value and should be differentiated from "not present". This PR fixes it so that they are set conditionally and only when present in the incoming YAML. I need this to render accurate terminal-friendly schemas for input/output models in Restish help documentation.

I also added some tests for these cases and fixed a couple of examples where the incorrect YAML node was passed in.

I'm not sure how these fields are getting used/checked by https://github.com/daveshanley/vacuum so you might want to double check I didn't break anything there with this change :smile:

### DIFF
--- a/datamodel/high/base/schema.go
+++ b/datamodel/high/base/schema.go
@@ -93,9 +93,15 @@ func NewSchema(schema *base.Schema) *Schema {
 	s := new(Schema)
 	s.low = schema
 	s.Title = schema.Title.Value
-	s.MultipleOf = &schema.MultipleOf.Value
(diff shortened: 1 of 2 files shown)


daveshanley (approved):
Looks good to me!

daveshanley:
Just an FYI, `vacuum` isn't using the models from libopenapi, it's using the index and the resolver however (but this change won't have any impact).

Thank you for your contribution, Daniel!

PR #21, opened 2022-11-29
fix(high): load schema.Examples into []any

danielgtaylor:
It looks like the `schema.Examples` field was missed when loading the high-level schema from a low-level one. This adds it along with a test. It allocates a new slice because we want to go from the low-level model to just the values in a `[]any`.


daveshanley (approved):
LGTM!

daveshanley:
Nice! I knew there would be a few of these missing in places. Thank you for your contribution!

PR #39, opened 2022-12-07
Bubble up circular reference errors

wdullaer:
First of: thanks for the library, it is an amazing parser.

I got here through restish, which recently started using libopenapi.
The issue I'm trying to solve here is that our APIs have circular references, but we don't really consider those to be errors.

The change here is to enable the client to make a choice whether circular references are errors or not: in case of circular references `BuildVxModel` will return both the list of errors and the Model.

An alternative implementation could be to have `BuildVxModel` take a boolean parameter `allowCircularReferences` which will prevent the circular reference check from happening. This feels more elegant to me, but is more disruptive for existing clients, so I chose not to go this way.

### DIFF
--- a/document.go
+++ b/document.go
@@ -15,11 +15,13 @@ package libopenapi
 
 import (
 	"fmt"
+
 	"github.com/pb33f/libopenapi/datamodel"
 	v2high "github.com/pb33f/libopenapi/datamodel/high/v2"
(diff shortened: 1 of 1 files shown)


daveshanley:
Thanks very much for your contribution and the compliments!

I'm happy with this implementation, it may need updating if other checks are required down the road, but it works in a non-breaking way. Thank you for helping make `libopenapi` better.

Issue #28, opened 2022-12-01
Items in Schema should be a single SchemaProxy instead of an array


TristanSpeakEasy:
https://json-schema.org/understanding-json-schema/reference/array.html

Items can either be a single schema (for v2, v3.0 and v3.1) or a boolean (v3.1).

And along with PrefixItems there should be a Contains field as well which is a single schema along with minContains / maxContains fields.

daveshanley:
Yes, this is a mistake on my part. It was during some copy pasta work. Good call on the missing JSON Schema properties ( I expect there are a few others missing also).

I'll look into this as soon as possible.

daveshanley:
I've changed this in the high level model so `Items` is a single `SchemaProxy` and added `contains` and the min/max properties also. However when I was thinking about the `or a boolean` part, I hit a crossroad and i'd like some of your input.

In order to handle these dynamic values, in the low level model I used a `SchemaDynamicValue` that can hold two types (A or B)
https://github.com/pb33f/libopenapi/blob/main/datamodel/low/base/schema.go#L15

This API is awkward, and there is a reason I didn't bring it up high, but it allows multiple types to sit in a single value across versions. The question I have is:

** Would this API be appropriate up high?** because to use a single property, the model would need to use this dynamic type approach or use interfaces.

Another question: ** What kind of signature/interface/design would you like to see?**

I have the rest of the code ready in a branch: 
https://github.com/pb33f/libopenapi/blob/issue28-contains/datamodel/high/base/schema.go#L58

Let me know what you think.

TristanSpeakEasy:
So kin-openapi as an example adds an additional field to the structure when a field can be represented by multiple types, here is an example for AdditionalProperties `https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L165` which could be one option but does mean the model doesn't match the spec exactly.

Or a container type that has a .Type() method which then allows you to case the .Value (any type) to the correct type to access the value?

daveshanley:
I'm open to suggestions on this one. Personally, it makes more sense to me to create a MultiType container (which is what the low-level does), but it might be too ugly for a high-level API.

Which of the two designs do you prefer? Another property or a multi-type container?

@danielgtaylor, what do you think?

TristanSpeakEasy:
I don't mind working with a container if it helps reduce some of the boiler plate around casting to types, maybe taking advantage of generics to return the correct type from a get method or similar

danielgtaylor:
[...]

PR #40, opened 2022-12-08
v0.4.0 Fixes #38 and #28 and introduces more 3.1 schema support.

### LINKED ISSUE #38: Schema `not` should be a single proxy instead of array
Similar to #28, it looks like `not` should be a single item instead of an array.

- https://json-schema.org/understanding-json-schema/reference/combining.html#not
- https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.2.1.4
- https://github.com/pb33f/libopenapi/blob/main/datamodel/high/base/schema.go#L59

It also looks like maybe conditional subschemas via [`if`, `then`, `else`](https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.2.2) are missing as well, but :scream: :skull:...

daveshanley:
`v0.4.0` Updates the `base.Schema` for both high and low-level models to fix some inaccuracies and introduce missing properties for JSON Schema / v3.1

## What was added

- `if`
- `else`
- `then`
- `propertyNames`
- `contains`
- `minContains`
- `maxContains`
- `unevaluatedItems`
- `unevaluatedProperties`
- `dependentSchemas`
- `patternProperties`

## What changed

`DynamicValue` was added as a container for values with differing types. This follows the same pattern established in the low level model.

**`Items`** can now be a `SchemaProxy` or a `bool` (using `DynamicValue`), no longer a slice.

**`Not`** is also no longer a slice.

**`ExclusiveMinimum`** and **`ExclusiveMaximum`** are now `DynamicValue` pointers that can be either a `bool` or `int64` 

The `ExclusiveMinimumBool` and `ExclusiveMaximumBool` properties have been removed as they are now combined.

These are breaking changes, but it does set a standard for how to proceed.


TristanSpeakEasy:
LGTM

danielgtaylor:
LGTM too, I'll take some time to upgrade Restish soon.

PR #51, opened 2022-12-15
feat: add path to references and parameters

TristanSpeakEasy:
This attempts to improve the completion of the index.Reference types when dealing with parameters so the path is available for rules in vacuum to use, as it is quite hard otherwise to determine the index of a parameter for a json path when using the index

The Official Go extension for VSCode reformatted the code, let me know if you want the formatting reverted


TristanSpeakEasy:
have also just added a fix for a duplicate reference in a parameters list

daveshanley (approved):
Looks good to me!

PR #55, opened 2022-12-20
feat: add an index for object schemas with properties and add additional info to enum reference

TristanSpeakEasy:


daveshanley (approved):
LGTM

PR #57, opened 2022-12-21
Fix setting JSON schema for OAS 3.1.0

hugoboos:
This fixes setting the correct JSON schema for OAS 3.1 documents. It was set to the default 3.0.


daveshanley (approved):
LGTM! Thank you for your contribution

Issue #49, opened 2022-12-14
Circular References may be incorrectly handled


TristanSpeakEasy:
In this document blob:https://docs.attentive.com/b6583e28-b7cf-40d0-a044-b6eef49eff3a

I get a circular reference error:

```
Circular reference detected: OrderProductModifier: Order -> OrderProduct -> OrderProductModifier -> OrderProductModifierCustomField [11250:31]
```

Looking into the document the actual circular reference seems to be `Order -> OrderProduct -> OrderProductModifier -> OrderProductModifier` through the `modifiers` field on `line 11287`

But it is listing `OrderProductModifierCustomField` as being where the circular reference is occurring but that is a simple object with no references in it.

Also from my understanding this is a "valid" circular reference as the `modifiers` field isn't a `required` field so the circular chain will break at some point (when that field is unset in the next level down from what we understand). So it won't infinitely be populated.

My understanding of this relationship between fields that cause circular references and `required` fields comes from this comment https://github.com/OAI/OpenAPI-Specification/issues/822#issuecomment-386937679

daveshanley:
I can't access the attentive link, I get a 404

TristanSpeakEasy:
@daveshanley make sure to include the blob: prefix

TristanSpeakEasy:
On my second point as well, and after reading your blog post I could ignore some of the circular reference errors if I deemed them safe, but would be good to get some help with that from the lib, so in the case where the data won't be infinitely recursive would be good to tell it is good recursion vs bad

daveshanley:
![Screenshot_20221215_095354](https://user-images.githubusercontent.com/187345/207892663-0a7dace0-31fa-4e29-b5dd-ba5ccbd649f5.png)
What am I doing wrong?

daveshanley:
> On my second point as well, and after reading your blog post I could ignore some of the circular reference errors if I deemed them safe, but would be good to get some help with that from the lib, so in the case where the data won't be infinitely recursive would be good to tell it is good recursion vs bad

What would you expect to see from the lib? What would you want to know? How would you want it to be exposed?

TristanSpeakEasy:
apologies seems it is a short lived URL, grab it from here by clicking the download button https://docs.attentive.com/openapi/reference/overview/

> What would you expect to see from the lib? What would you want to know? How would you want it to be exposed?

Maybe the recursive error that is returned contains a flag that is set to true when the recursion isn't infinite?

daveshanley:
[...]

PR #64, opened 2023-01-05
feat: Allow circular references which contain optional properties in their journeys

BenjaminNolan:
This PR contains an initial fix for pb33f/libopenapi#49.

It adds the following features to the library:

* `index.CircularReferenceResult.IsInfiniteLoop` - Set to true if each definition in the journey is marked as required
* `index.Reference.RequiredRefProperties` - Contains a map of definition references to the property name(s) which use it

`resolver.Resolver.CheckForCircularReferences()` and `resolver.Resolver.Resolve()` now check for `circRef.IsInfiniteLoop` when iterating through `resolver.Resolver.circularReferences`. If the value is `false`, then this is now accepted and no error is added to `resolver.Resolver.resolvingErrors`, as optional items within circular references allow a termination point, thus it's a valid spec.

### Remote and file refs

I noticed whilst reading through the code that it will automatically load references from remote URLs should they exist in the specs (eg, as shown in `mixedref-burgershop.openapi.yaml`. This is a potential security hole, and should be switched off by default, exposed via an option either passed in to `resolver.NewResolver()`, or set via a receiver function, eg `resolver.Resolver.SetAllowRemoteReferences(true)`. This should probably also be done for file refs as well. The fiasco with `log4j` ([CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228)) happened because exactly this sort of on-by-default behaviour, so I think we should learn from their mistake!

### Option to treat errors as warnings

Currently, circular references where the whole chain is marked as required cause a hard error, and the spec doesn't import, however they _are_ theoretically loadable since subschemas are only built on request. I've tested this by commenting out the loops in `resolver.Resolver.Resolve()` and `resolver.Resolver.CheckForCircularReferences()`, and the Stripe spec successfully imports into our system.

## Tests

I've added two additional files to `test_specs`:
* `swagger-invalid-recursive-model.yaml` - Contains a minimal *invalid* Swagger 2.0 example (based on the Shortcut spec) where each child reference is marked as required
* `swagger-valid-recursive-model.yaml` - Contains a minimal *valid* Swagger 2.0 example (also based on the Shortcut spec) where each child reference is optional

The Stripe test was expecting 23 errors, however 20 of these were terminable loops, thus are now valid. The remaining three are:

```plain
#/components/schemas/payment_intent
  -> #/components/schemas/payment_intent
  -> #/components/schemas/customer
  -> #/components/schemas/invoice_setting_customer_setting
  -> #/components/schemas/payment_method
  -> #/components/schemas/payment_method_card
  -> #/components/schemas/payment_method_card_generated_card
  -> #/components/schemas/setup_attempt
  -> #/components/schemas/api_errors
  -> #/components/schemas/payment_intent
```

```plain
#/components/schemas/balance_transaction
  -> #/components/schemas/dispute
  -> #/components/schemas/balance_transaction
```

```plain
#/components/schemas/transfer_reversal
  -> #/components/schemas/transfer
  -> #/components/schemas/transfer_reversal
```

Thus, I've updated the Stripe test to expect these three invalid circular references, as the others are all valid since they can all terminate at least once in the chain.

### TestResolver_ResolveComponents_MixedRef

This test is currently failing in `main`. I've added a couple of TODOs around it relating to the remote refs note above, and the line which is causing the failing test. I did have a look into what's going on, but I'm not sure what the expected 10 errors would've been, so have left it in its current state.

Shoutout to @taylow for helping with this. :)

### DIFF
--- a/datamodel/low/reference_test.go
+++ b/datamodel/low/reference_test.go
@@ -133,8 +133,40 @@ func TestIsCircular_LookupFromJourney(t *testing.T) {
 	assert.True(t, IsCircular(ref, idx))
 }
 
-func TestIsCircular_LookupFromLoopPoint(t *testing.T) {
+func TestIsCircular_LookupFromJourney_Optional(t *testing.T) {
 
+	yml := `components:
+  schemas:
+    Something:
+      properties:
+        nothing:
+          $ref: '#/components/schemas/Nothing'
+    Nothing:
+      properties:
+        something: 
+          $ref: '#/components/schemas/Something'
+`
+
+	var iNode yaml.Node
+	mErr := yaml.Unmarshal([]byte(yml), &iNode)
+	assert.NoError(t, mErr)
+	idx := index.NewSpecIndex(&iNode)
+
+	yml = `$ref: '#/components/schemas/Something'`
+
+	resolve := resolver.NewResolver(idx)
+	errs := resolve.CheckForCircularReferences()
+	assert.Len(t, errs, 0)
+
+	var idxNode yaml.Node
+	_ = yaml.Unmarshal([]byte(yml), &idxNode)
+
+	ref, err := LocateRefNode(idxNode.Content[0], idx)
+	assert.NoError(t, err)
+	assert.True(t, IsCircular(ref, idx))
+}
+
+func TestIsCircular_LookupFromLoopPoint(t *testing.T) {
 	yml := `components:
   schemas:
     Something:
@@ -161,6 +193,38 @@ func TestIsCircular_LookupFromLoopPoint(t *testing.T) {
 	resolve := resolver.NewResolver(idx)
 	errs := resolve.CheckForCircularReferences()
 	assert.Len(t, errs, 1)
+
+	var idxNode yaml.Node
+	_ = yaml.Unmarshal([]byte(yml), &idxNode)
+
+	ref, err := LocateRefNode(idxNode.Content[0], idx)
+	assert.NoError(t, err)
+	assert.True(t, IsCircular(ref, idx))
+}
+
+func TestIsCircular_LookupFromLoopPoint_Optional(t *testing.T) {
+	yml := `components:
+  schemas:
+    Something:
+      properties:
+        nothing:
+          $ref: '#/components/schemas/Nothing'
+    Nothing:
+      properties:
+        something: 
+          $ref: '#/components/schemas/Something'
+`
+
+	var iNode yaml.Node
+	mErr := yaml.Unmarshal([]byte(yml), &iNode)
+	assert.NoError(t, mErr)
+	idx := index.NewSpecIndex(&iNode)
+
+	yml = `$ref: '#/components/schemas/Nothing'`
+
+	resolve := resolver.NewResolver(idx)
+	errs := resolve.CheckForCircularReferences()
+	assert.Len(t, errs, 0)
 
 	var idxNode yaml.Node
 	_ = yaml.Unmarshal([]byte(yml), &idxNode)
@@ -211,6 +275,42 @@ func TestIsCircular_FromRefLookup(t *testing.T) {
 	assert.False(t, IsCircular(idxNode.Content[0], idx))
 }
 
+func TestIsCircular_FromRefLookup_Optional(t *testing.T) {
+	yml := `components:
+  schemas:
+    NotCircle:
+      description: not a circle
+    Something:
+      properties:
+        nothing:
+          $ref: '#/components/schemas/Nothing'
+    Nothing:
+      properties:
+        something: 
+          $ref: '#/components/schemas/Something'
+`
+
+	var iNode yaml.Node
+	mErr := yaml.Unmarshal([]byte(yml), &iNode)
+	assert.NoError(t, mErr)
+	idx := index.NewSpecIndex(&iNode)
+
+	resolve := resolver.NewResolver(idx)
+	errs := resolve.CheckForCircularReferences()
+	assert.Len(t, errs, 0)
+
+	yml = `$ref: '#/components/schemas/Nothing'`
+	var idxNode yaml.Node
+	_ = yaml.Unmarshal([]byte(yml), &idxNode)
+
+	assert.True(t, IsCircular(idxNode.Content[0], idx))
+
+	yml = `$ref: '#/components/schemas/NotCircle'`
+	_ = yaml.Unmarshal([]byte(yml), &idxNode)
+
+	assert.False(t, IsCircular(idxNode.Content[0], idx))
+}
+
 func TestIsCircular_NoNode(t *testing.T) {
 	assert.False(t, IsCircular(nil, nil))
 }
@@ -220,7 +320,6 @@ func TestGetCircularReferenceResult_NoNode(t *testing.T) {
 }
 
 func TestGetCircularReferenceResult_FromJourney(t *testing.T) {
-
 	yml := `components:
   schemas:
     Something:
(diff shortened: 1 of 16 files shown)


index/spec_index.go
```diff
@@ -1698,18 +1701,93 @@ func (index *SpecIndex) extractDefinitionsAndSchemas(schemasNode *yaml.Node, pat
...
 	}
 }
 
+// extractDefinitionRequiredRefProperties goes through the direct properties of a schema and extracts the map of required definitions from within it
+func (index *SpecIndex) extractDefinitionRequiredRefProperties(schemaNode *yaml.Node, reqRefProps map[string][]string) map[string][]string {
+	if schemaNode == nil {
+		return reqRefProps
+	}
+
+	_, requiredSeqNode := utils.FindKeyNode("required", schemaNode.Content)
```
daveshanley, line 1724: I would use `utils.FindKeyNodeTop` instead of `utils.FindKeyNode`. The latter will look through two levels for the key, vs just the top. It can give inaccurate results. and find child keys.

index/spec_index.go
```diff
@@ -1698,18 +1701,93 @@ func (index *SpecIndex) extractDefinitionsAndSchemas(schemasNode *yaml.Node, pat
...
+	if requiredSeqNode == nil {
+		return reqRefProps
+	}
+
+	_, propertiesMapNode := utils.FindKeyNode("properties", schemaNode.Content)
```
daveshanley, line 1729: Same as above with ` utils.FindKeyNodeTop`

index/spec_index.go
```diff
@@ -1698,18 +1701,93 @@ func (index *SpecIndex) extractDefinitionsAndSchemas(schemasNode *yaml.Node, pat
...
+			if ofNode != nil {
+				for _, ofNodeItem := range ofNode.Content {
+					reqRefProps = index.extractRequiredReferenceProperties(ofNodeItem, name, reqRefProps)
+				}
+			}
+		}
+	}
+
+	for _, requiredPropertyNode := range requiredSeqNode.Content {
+		_, requiredPropDefNode := utils.FindKeyNode(requiredPropertyNode.Value, propertiesMapNode.Content)
```
daveshanley, line 1758: Same as above with `utils.FindKeyNodeTop`

index/spec_index.go
```diff
@@ -1698,18 +1701,93 @@ func (index *SpecIndex) extractDefinitionsAndSchemas(schemasNode *yaml.Node, pat
...
+	}
+
+	return reqRefProps
+}
+
+// extractRequiredReferenceProperties returns a map of definition names to the property or properties which reference it within a node
+func (index *SpecIndex) extractRequiredReferenceProperties(requiredPropDefNode *yaml.Node, propName string, reqRefProps map[string][]string) map[string][]string {
+	isRef, _, defPath := utils.IsNodeRefValue(requiredPropDefNode)
+	if !isRef {
+		_, defItems := utils.FindKeyNode("items", requiredPropDefNode.Content)
```
daveshanley, line 1773: Same as above with `utils.FindKeyNodeTop`

utils/utils.go
```diff
@@ -402,15 +403,50 @@ func IsNodeRefValue(node *yaml.Node) (bool, *yaml.Node, string) {
 	return false, nil, ""
 }
 
+// IsPropertyNodeRequired will check if a node is required within circular references
+func IsPropertyNodeRequired(node *yaml.Node, propertyName string) bool {
+	_, requiredSeqNode := FindKeyNode("required", node.Content)
```
daveshanley, line 408: Same as above with `utils.FindKeyNodeTop`

utils/utils.go
```diff
@@ -402,15 +403,50 @@ func IsNodeRefValue(node *yaml.Node) (bool, *yaml.Node, string) {
...
+		return false
+	}
+
+	_, propertiesMapNode := FindKeyNode("properties", node.Content)
+	if propertiesMapNode == nil {
+		return false
+	}
+
+	for _, requiredPropertyNode := range requiredSeqNode.Content {
+		_, requiredPropDefNode := FindKeyNode(requiredPropertyNode.Value, propertiesMapNode.Content)
```
daveshanley, line 419: Same as above with `utils.FindKeyNodeTop`

utils/utils.go
```diff
@@ -402,15 +403,50 @@ func IsNodeRefValue(node *yaml.Node) (bool, *yaml.Node, string) {
...
+		if requiredPropDefNode == nil {
+			continue
+		}
+
+		isRef, _, defPath := IsNodeRefValue(requiredPropDefNode)
+		if isRef && defPath == propertyName {
+			return true
+		}
+
+		_, defItems := FindKeyNode("items", requiredPropDefNode.Content)
```
daveshanley, line 429: Same as above with `utils.FindKeyNodeTop`

daveshanley (commented):
Wow, this is great, a huge contribution! Thank you for taking this on, it's a real challenge!

index/spec_index.go
BenjaminNolan: Cool, I've updated the PR! :)

daveshanley:
I see the tests are throwing a stack overflow. I can't tell you how many nights this has robbed from me when building the mechanics to prevent run-away recursion :)

BenjaminNolan:
Hmm, I was getting this intermittently in the tests, but wasn't sure why. It seemed to work most of the time, and only occasionally SO. I suspect it's down to the fact that Go picks random orders for items in maps, so when it succeeds it's loaded the items in an order which avoids the overflow. I have been testing the import via our actual app with this version of the code, and not once has it even gone high with memory usage there.

taylow:
It seems like the ordering within the `RequiredRefProperties` map being iterated over inside the recursive `isInfiniteCircularDependency` func is causing the intermittent behaviour, as this does pass sometimes.

Perhaps a hard limit could be used in `isInfiniteCircularDependency` that determines how many levels is considered "infinite". This could be exposed as a parameter/option on the library, that way you pass on the decision of recursive depth to the client and they can make the decision.

I believe Ben added TODOs regarding remote schemas and Log4J (different problem, but similar possible attack vector). It seems like a few options on the library wouldn't go amiss!

daveshanley:
Adding configuration params would be pretty straightforward from a design perspective - it would make sense. There are some design considerations on the API at the top of the library, a new `CreateDocumentFromConfig` function would do the trick.

BenjaminNolan:
I've added a 'visited' check to `isInfiniteCircularReference()` which has fixed the stack overflow.

I've also done my best to ensure that the existing behaviour is retained as much as possible. The following structure defines an invalid schema loop, as each of the entries indicates that its content is defined in another schema. I've added a check to see if the top-level schema is a direct reference, and we now treat it as required.

```
components:
  schemas:
    loopy:
      $ref: '#/components/schemas/cake'
    cake:
      $ref: '#/components/schemas/loopy'
    pizza:
      $ref: '#/components/schemas/cake'
```

The tests should all pass correctly now, and the behaviour of the library should be the same save for properly handling terminable circular references. :)

We need the facility to be able to import invalid specs and handle them as much as possible, so I'll look at adding `CreateDocumentFromConfig` next.

datamodel/low/reference_test.go
```diff
@@ -281,10 +307,15 @@ func TestGetCircularReferenceResult_FromMappedRef(t *testing.T) {
       properties:
         nothing:
           $ref: '#/components/schemas/Nothing'
+      required:
+        - nothing
     Nothing:
       properties:
         something: 
-          $ref: '#/components/schemas/Something'`
+          $ref: '#/components/schemas/Something'
```
daveshanley, line 315: All the tests in here have `required` added. Is it possible to ensure that all this code works without `required` being present? I don't want to suddenly see a significant regression where schemas that have not used `required` no longer work.

daveshanley:
There is a drop in coverage of 0.32% - it's too high as it drops us below a minimum 99.5% threshold. It would drop the badge from 100% coverage to 99% coverage. Can we nudge it back up a few points?

(I have spent a lot of time maintaining high coverage to ensure as much accuracy as possible)

datamodel/low/reference_test.go
BenjaminNolan: So, the reason I added the `required` lists here is to ensure that the existing tests which detect infinite circular references retain their existing behaviour.

Without the `required` list entries, the circular references in these tests are now valid, as they can be terminated at least once during the chain, so the tests would fail without the actual test behaviour being modified.

BenjaminNolan:
I've removed an unused util function we added during the initial pass on this feature which we didn't end up using, but left in as it struck me as potentially useful to someone else, as it was this function which was dropping the code coverage below 99.5%. I've also added a test for `resolver.ResolvingError.Error()` which wasn't previously being tested to bump it up a bit more. :)

datamodel/low/reference_test.go
BenjaminNolan: I've added versions of each of these tests for their equivalents without the required flags, which should ensure that both sides of the equation work. :)

daveshanley (approved):
Thank you for your contribution and for helping make this a better library.

BenjaminNolan:
You're welcome, am looking forward to the next one! ^^

PR #67, opened 2023-01-12
Fix ref object parsing

bitomaxsp:
If schemas item is a `$ref` then it wasn't parsed using BuildModel.

This PR moves BuildModel to lo schema.Build and works on possible dereferenced root yaml node.


bitomaxsp:
Hi @daveshanley . Can you take a look into that and merge if it's ok. It is kind of critical thing since some fields are not parsed if schema is a pure ref to another schema. Thanks :)

daveshanley (approved):
LGTM! Thank you for your contribution!

Issue #70, opened 2023-01-19
callbacks map empty in v3.Operation


TristanSpeakEasy:
Trying to iterate over callbacks for an operation but the map is always empty.

Here is a reproducible

```go
package main

import (
	"fmt"

	"github.com/pb33f/libopenapi"
)

func main() {
	doc, err := libopenapi.NewDocument([]byte(`openapi: 3.1.0
info:
  title: Test
  version: 0.0.4
  description: A test API
servers:
  - url: https://example.com
paths:
  /test:
    get:
      operationId: test
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  message:
                    type: string
                    example: Hello World
      callbacks:
        newPet:
          '{$request.body#/callbackUrl}':
            post:
              requestBody:
                description: Information about a new pet in the system
                content:
                  application/json:
                    schema:
                      $ref: '#/components/schemas/Pet'
              responses:
                '200':
                  description: Return a 200 status to indicate that the data was received successfully
components:
  schemas:
    Pet:
      type: object
      properties:
        name:
          type: string
        tag:
          type: string
`))
	if err != nil {
		panic(err)
	}

	m, errs := doc.BuildV3Model()
	if errs != nil {
		panic(errs)
	}

	fmt.Println(m.Model.Paths.PathItems["/test"].Get.Callbacks)
}
```

TristanSpeakEasy:
Can confirm this is now fixed

PR #71, opened 2023-01-20
fix: callbacks not being populated in high model

TristanSpeakEasy:
addresses https://github.com/pb33f/libopenapi/issues/70


daveshanley (approved):
Thank you for finding and fixing this! Guess I missed a bit :)

PR #72, opened 2023-01-21
fix: DynamicValue in high model schema missing N value

danielgtaylor:
Found a small issue while upgrading to the latest libopenapi. The new `DynamicValue.N` field is not copied over from the low-level model, so for example `schema.ExclusiveMinimum.IsB()` will *always* return `false`. This PR fixes it and adds a small test to assert the values get set. For the boolean case it is left alone as `N` will default to zero.

Note: this is important as the exclusive minimum can be zero, in which case it's currently not possible to determine whether a value was set or which one was set.

Sorry for the delay in upgrading BTW, I got side-tracked shipping https://rest.sh/#/bulk


TristanSpeakEasy (approved):

daveshanley:
Thank you for fixing it and for your contribution!

PR #75, opened 2023-02-01
fix: for low model reference values in arrays being set incorrectly

TristanSpeakEasy:
@daveshanley I found a bug where in an array of parameters (example below) if the array had some references and some not then the values in the array without references were being set to the last found reference which came from the previous value in the array.

I fixed this for the use case I found, but not sure if you want to have a look over this to see if there is anywhere else this behavior might happen.

would also appreciate help putting together a test for this fix, in my rush I couldn't get my head around the best way to write a test for this.

here is the example that this fixes:

```yaml
parameters:
        - $ref: "#/components/parameters/filterStart"
        - $ref: "#/components/parameters/filterLimit"
        - $ref: "#/components/parameters/filterEnd"
        - $ref: "#/components/parameters/filterDirection"
        - description: Specifies the unit of aggregation in the returned results.
          in: query
          name: unit
          schema:
            default: minute
            enum:
              - minute
              - hour
              - day
              - month
            type: string
```

you can find it in this document here: https://github.com/speakeasy-api/openapi-directory/blob/main/APIs/ably.io/platform/1.1.0/openapi.yaml

### DIFF
--- a/datamodel/low/extraction_functions.go
+++ b/datamodel/low/extraction_functions.go
@@ -6,12 +6,13 @@ package low
 import (
 	"crypto/sha256"
 	"fmt"
+	"strconv"
+	"strings"
+
 	"github.com/pb33f/libopenapi/index"
 	"github.com/pb33f/libopenapi/utils"
 	"github.com/vmware-labs/yaml-jsonpath/pkg/yamlpath"
 	"gopkg.in/yaml.v3"
-	"strconv"
-	"strings"
 )
 
 // FindItemInMap accepts a string key and a collection of KeyReference[string] and ValueReference[T]. Every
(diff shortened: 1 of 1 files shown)


datamodel/low/extraction_functions.go
```diff
@@ -287,11 +293,17 @@ func ExtractArray[T Buildable[N], N any](label string, root *yaml.Node, idx *ind
...
+				localReferenceValue = referenceValue
+				localIsReference = isReference
+			}
+
 			items = append(items, ValueReference[T]{
 				Value:       n,
 				ValueNode:   node,
-				IsReference: isReference,
-				Reference:   referenceValue,
+				IsReference: localIsReference,
```
TristanSpeakEasy, line 305: this is the fix, apologies for gofmt changing so much, I can revert the other changes if you want

daveshanley (approved):
I regard to a test for this, I am not entirely sure how to write it yet, just by reading the code and looking at the spec. We may just need to document it in the code somewhere, the coverage hasn't dropped any.

Issue #73, opened 2023-01-21
Unable to resolve relative URL references


danielgtaylor:
I love the DigitalOcean OpenAPI spec ([raw](https://raw.githubusercontent.com/digitalocean/openapi/main/specification/DigitalOcean-public.v2.yaml) | [github](https://github.com/digitalocean/openapi/tree/main/specification)) for its ability to break tools: loads of relative references and even the text fields like `description` use `$ref`s to separate YAML files containing Markdown prose. It seems like libopenapi can't load it when loading from a URL due to the references:

```sh
$ restish digitalocean --help
panic: errors [path item build failed: cannot find reference: resources/1-clicks/oneClicks_install_kubernetes.yml at line 539, col 13]
```

That makes sense, because when creating a new `libopenapi.Document` you don't pass in a base URI, just the loaded bytes of the YAML. The [code to determine whether some reference should be loaded via HTTP](https://github.com/pb33f/libopenapi/blob/main/index/spec_index.go#L1684) is run much later down the chain:

```go
func DetermineReferenceResolveType(ref string) int {
	if ref != "" && ref[0] == '#' {
		return LocalResolve
	}
	if ref != "" && len(ref) >= 5 && (ref[:5] == "https" || ref[:5] == "http:") {
		return HttpResolve
	}
	if strings.Contains(ref, ".json") ||
		strings.Contains(ref, ".yaml") ||
		strings.Contains(ref, ".yml") {
		return FileResolve
	}
	return -1
}
```

From the [OpenAPI 3.1.0 spec](https://spec.openapis.org/oas/v3.1.0#relative-references-in-uris):

> Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [[RFC3986](https://spec.openapis.org/oas/v3.1.0#bib-RFC3986)].
> 
> Relative references, including those in [Reference Objects](https://spec.openapis.org/oas/v3.1.0#referenceObject), [PathItem Object](https://spec.openapis.org/oas/v3.1.0#pathItemObject) $ref fields, [Link Object](https://spec.openapis.org/oas/v3.1.0#linkObject) operationRef fields and [Example Object](https://spec.openapis.org/oas/v3.1.0#exampleObject) externalValue fields, are resolved using the referring document as the Base URI according to [[RFC3986](https://spec.openapis.org/oas/v3.1.0#bib-RFC3986)].

This doesn't seem like a quick fix, so I wanted to open this issue to discuss how to approach it.

[...]

PR #83, opened 2023-02-18
v0.6.0 Brings support for relative URLs and paths. #73

daveshanley:
`v0.6.0` Addresses issue #73 by introducing some enhancements to the index, allowing for relative file handling, and providing a fix to auto-allowing remote and local files to be followed.

A new configuration option is available to document creation for an Index. It's defined as `index.SpecIndexConfig` and provides three properties: 

- `BaseURL` of type `*url.URL`
- `AllowRemoteLookup` of type `bool`
- `AllowFileLookup` of type `bool`

This can be used to configure an index to know where to look when encountering relative paths and if to allow them at all.

[Full documentation can be found here](https://pb33f.io/libopenapi/index/#configuration).

When creating a new `Document`, there is also a new `datamodel.DocumentConfigutation`, which looks almost identical, except
for the **Lookup** postfix has been replaced with **References**

```
config := datamodel.DocumentConfiguration{
		AllowFileReferences:   true,
		AllowRemoteReferences: true,
		BaseURL:               baseURL,
	}
```

This new config can be used with a new function called `NewDocumentWithConfiguration()` that is the same as `NewDocument()` except it has a second argument for accepting the config.

```
doc, err := NewDocumentWithConfiguration(digitalOcean, &config)
```

[Full documentation can be found here](https://pb33f.io/libopenapi/openapi/#documents-with-relative-or-remote-references).

The index has been refactored a little to help make it easier to navigate and tuned up to run faster when indexing references.

The digital ocean spec has been added for testing purposes.


index/find_component.go
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+
+        d := make(chan bool)
+        var body []byte
+        var err error
+
+        go func(uri string) {
+            bc := make(chan []byte)
+            ec := make(chan error)
+            go getRemoteDoc(uri, bc, ec)
```
n0rig, line 122: wow, nice.

index/find_component_test.go
```diff
@@ -0,0 +1,101 @@
...
+    var rootNode yaml.Node
+    _ = yaml.Unmarshal([]byte(yml), &rootNode)
+
+    c := CreateOpenAPIIndexConfig()
+    index := NewSpecIndexWithConfig(&rootNode, c)
+
+    thing := index.FindComponentInRoot("#/valid-but-missing")
+    assert.Nil(t, thing)
+    assert.Len(t, index.GetReferenceIndexErrors(), 1)
+}
```
n0rig, line 101: Can you add a test case for path traversal attacks.
e.g. 
* `https://pb33f.io/../../../etc/passwd`

index/index_model.go
```diff
@@ -0,0 +1,258 @@
...
+        AllowRemoteLookup: false,
+        AllowFileLookup:   false,
+        seenRemoteSources: &syncmap.Map{},
+    }
+}
+
+// SpecIndex is a complete pre-computed index of the entire specification. Numbers are pre-calculated and
+// quick direct access to paths, operations, tags are all available. No need to walk the entire node tree in rules,
+// everything is pre-walked if you need it.
+type SpecIndex struct {
```
n0rig, line 100: legit; so, this the holy grail of the entire spec.

n0rig: imagine allowing caching this in datastore (s3) in the future

index/index_utils.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+    "gopkg.in/yaml.v3"
+    "net/http"
+    "strings"
+    "time"
+)
+
+func isHttpMethod(val string) bool {
+    switch strings.ToLower(val) {
```
n0rig, line 14: nit: kind of unreadable; I guess these match up to [these](https://github.com/pb33f/libopenapi/blob/7c6507488d9734b961d2d49514fefdb5350c8359/index/index_model.go#L258)?

resolver/resolver.go
```diff
@@ -93,38 +96,33 @@ func (resolver *Resolver) GetNonPolymorphicCircularErrors() []*index.CircularRef
...
+func (resolver *Resolver) GetReferenceVisited() int {
+	return resolver.referencesVisited
+}
+
+// GetIndexesVisited returns the number of indexes visited by the resolver
+func (resolver *Resolver) GetIndexesVisited() int {
+	return resolver.indexesVisited
+}
+
+//
```
n0rig, line 114: Missing comment

resolver/resolver.go
```diff
@@ -179,8 +161,71 @@ func (resolver *Resolver) CheckForCircularReferences() []*ResolvingError {
...
+			res.journeysTaken++
+			res.VisitReference(schemaRef, seenReferences, journey, false)
+		}
+	}
+	for _, c := range idx.GetChildren() {
+		visitIndexWithoutDamagingIt(res, c)
+	}
+}
+
+func visitIndex(res *Resolver, idx *index.SpecIndex) {
```
n0rig, line 188: why does this damage it? (versus the function above `visitIndexWithoutDamagingIt`)

n0rig (changes requested):
take or leave it; most of these comments are nits. I think some would be nice to address.

Otherwise, the only other comment is that maybe the index code could be brought in as another PR.

datamodel/document_config.go
```diff
@@ -0,0 +1,19 @@
...
+import "net/url"
+
+// DocumentConfiguration is used to configure the document creation process. It was added in v0.6.0 to allow
+// for more fine-grained control over controls and new features.
+type DocumentConfiguration struct {
+    // if the document uses relative file references, this is the base url to use when resolving them.
+    BaseURL *url.URL
+
+    // AllowFileReferences will allow the index to locate relative file references. This is disabled by default.
+    AllowFileReferences bool
```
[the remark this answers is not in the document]
daveshanley: The thing is that files may not be relative, they may be absolute, for example:

`$ref: '/some/path/to/my/file.json#/a/component'`

and 

`$ref : '../../my/file.json#/a/component`

both need to be supported and we can't assume references will always be relative when used with files.

datamodel/low/v2/swagger.go
```diff
@@ -121,13 +121,32 @@ func (s *Swagger) GetExtensions() map[low.KeyReference[string]]low.ValueReferenc
...
+    return createDocument(info, configuration)
+}
+
+// CreateDocument will create a new Swagger document from the provided SpecInfo.
+//
+// Deprecated: Use CreateDocumentFromConfig instead.
 func CreateDocument(info *datamodel.SpecInfo) (*Swagger, []error) {
+    return createDocument(info, &datamodel.DocumentConfiguration{
+        AllowRemoteReferences: true,
+        AllowFileReferences:   true,
```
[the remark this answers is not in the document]
daveshanley: No, the original behavior needs to be maintained for backward compatibility. There were no limitations before, so there can't be now.

document.go
```diff
@@ -39,6 +39,10 @@ type Document interface {
 	// GetSpecInfo will return the *datamodel.SpecInfo instance that contains all specification information.
 	GetSpecInfo() *datamodel.SpecInfo
 
+	// SetConfiguration will set the configuration for the document. This allows for finer grained control over
+	// allowing remote or local references, as well as a BaseURL to allow for relative file references.
+	SetConfiguration(configuration *datamodel.DocumentConfiguration)
```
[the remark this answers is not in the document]
daveshanley: I would agree if there were multiple configuration settings for a document, but there is only one, so I am not sure what clarity it adds?

datamodel/high/v3/document_test.go
```diff
@@ -398,6 +398,18 @@ func TestAsanaAsDoc(t *testing.T) {
 	fmt.Println(d)
 }
 
+func TestDigitalOceanAsDoc(t *testing.T) {
+	data, _ := ioutil.ReadFile("../../../test_specs/asana.yaml")
```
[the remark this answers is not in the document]
daveshanley: Err.. this is a mistake, and I will rectify, Nice catch.

document.go
```diff
@@ -120,23 +139,32 @@ func (d *document) BuildV2Model() (*DocumentModel[v2high.Swagger], []error) {
...
 		return nil, errors
 	}
-	lowDoc, errs := v2low.CreateDocument(d.info)
-	// Do not shortcircuit on circular reference errors, so the client
+
+	var lowDoc *v2low.Swagger
+	if d.config == nil {
+		d.config = &datamodel.DocumentConfiguration{
+			AllowFileReferences:   true,
+			AllowRemoteReferences: true,
```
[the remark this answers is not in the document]
daveshanley: Agreed, will bump comments.

document_test.go
```diff
@@ -247,6 +249,77 @@ func ExampleNewDocument_fromOpenAPI3Document() {
...
+func ExampleNewDocument_fromWithDocumentConfigurationSuccess() {
+
+	// This example shows how to create a document that prevents the loading of external references/
+	// from files or the network
+
+	// load in the Digital Ocean OpenAPI specification
+	digitalOcean, _ := ioutil.ReadFile("test_specs/digitalocean.yaml")
+
+	// Digital Ocean needs a baseURL to be set, so we can resolve relative references.
+	baseURL, _ := url.Parse("https://raw.githubusercontent.com/digitalocean/openapi/main/specification")
```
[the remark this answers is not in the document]
daveshanley: Done. Added to `datamodel/high/v3/document_test.go` Tests the baseURL from that exact SHA. Works as expected.

index/find_component.go
```diff
@@ -0,0 +1,337 @@
...
+        if index.config.AllowFileLookup {
+            return index.lookupFileReference(id)
+        } else {
+            return nil, nil, fmt.Errorf("local lookups are not permitted, " +
+                "please set AllowFileLookup to true in the configuration")
+        }
+    }
+
+    switch DetermineReferenceResolveType(componentId) {
+    case LocalResolve: // ideally, every single ref in every single spec is local. however, this is not the case.
```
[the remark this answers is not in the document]
daveshanley: This is part of the original codebase, `local` means `local to the spec`; anything else requires an external lookup, which makes it foreign or `non-local.`

In the context of the library, it makes perfect sense to me :)

In the context of an OS, yes, local means something different.

index/find_component_test.go
daveshanley: How? without a whitelist, what would the test be?

index/index_model.go
daveshanley: It looks like someone can see into the future!

index/index_utils.go
daveshanley: Old refactored code won't fix.

resolver/resolver.go
daveshanley: willfix

resolver/resolver.go
daveshanley: See: https://github.com/pb33f/libopenapi/blob/fix-73/resolver/resolver.go#L198

And: 

https://github.com/pb33f/libopenapi/blob/fix-73/resolver/resolver.go#L208

and

https://github.com/pb33f/libopenapi/blob/fix-73/resolver/resolver.go#L217

The resolver will destroy the root node by literally sticking together every reference found, inline, into a single giant mega tree. line numbers will be messed up, but the whole thing is walkable using code.

However the resolver isn't used to resolve by the datamodel, it's used by `vacuum`.

datamodel/low/v3/path_item.go
```diff
@@ -210,7 +210,7 @@ func (p *PathItem) Build(root *yaml.Node, idx *index.SpecIndex) error {
 					pathNode.Content[1].Value, pathNode.Content[1].Line, pathNode.Content[1].Column)
 			}
 		}
-
+		wg.Add(1)
```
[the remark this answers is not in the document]
daveshanley: A wait group is a simple helper for go routines that allows items of work to be added to the queue in the group (`Add(items)`), and then as each thread finishes an item of work, it calls  (`Done()`) to inform the group, there is one less thing in the queue.

When the queue is empty, the group is clear, and work can proceed beyond where it was paused. The pause allows an undetermined amount of go routines to fly away, do whatever they are doing, and then report back to a single place when they are done.

However, the problem with that blurb is that it's not relevant to the library.

resolver/resolver.go
n0rig: ack; that makes sense now.

n0rig (approved):

PR #98, opened 2023-03-17
fix: add missing summary for OpenAPI 3.X docs

TristanSpeakEasy:
This adds the missing Summary attribute to the doc.Info object for 3.X specs, this isn't present for OpenAPI 2 specs which means that it will always be empty but I think that is okay and avoided the need to have a v2 and v3 version of this object.

Also the TestSpecIndex_SearchIndexForReference_ExternalSpecs test was failing which I believe is unrelated to this change and it failed non-deterministically sometimes timing out all together


daveshanley:
Wow.. how did I miss that one! Thank you for fixing it.

PR #100, opened 2023-03-25
Fix multi file resolving

bitomaxsp:
Fix Digital Ocean resolving (invalid spec)
Fix valid specs resolving


daveshanley (approved):
Looks good to me, Thank you for your contribution.

Issue #121, opened 2023-05-30
Maximum and Minimum should be *float64


thrawn01:
I think the `Maximum` and `Minimum` values in `base.Schema` should be `float64`. As defined now it's impossible to define a float64 (aka number) as these values. 

[JSON Schema](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.2.2) defines these as `number` which probably means we need to represent them as `*float64` in golang structs.

daveshanley:
Available in `v0.9.0`

PR #122, opened 2023-05-30
fix: Schema.Minimum and Schema.Maxmium are now float64

thrawn01:
## Purpose
`Maximum` and `Minimum` values in `base.Schema` should be `float64`

[JSON Schema](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.2.2) defines these as `number` which means we need to represent them as `*float64` in golang structs.

## Implementation
* Introduced `utils.IsNodeNumberValue()` to validate if the YAML value can be safely parsed as a float64.
* `ExclusiveMaximum` now accepts a float64 (number)
* `ExclusiveMinimum` now accepts a float64 (number)
* `Maximum` now accepts a float64 (number)
* `Minimum` now accepts a float64 (number)
* `MultipleOf` now accepts a float64 (number)

### Versioning
Recommend holding this PR until library is ready to release `v1` or we could optionally tag `v1.0.0-rc1`until a full `v1` release is ready.

### PostScript
I noticed too late that all the files use white space as indentation, when I ran `go fmt` after making my changed (which is standard ) I found to my horror that every line in every file had changed. At that point my options were attempt to undo all the white space changes `go fmt` had preformed or wipe out all my changes and redo the Pull Request. I opted to re-write the PR. This was mildly infuriating and may trip up other contributors. Coming from python myself, I felt the frustration of tabs when moving to golang, but I've since embraced it as `go fmt` is such a nice tool.

Reference: https://github.com/pb33f/libopenapi/issues/121


utils/utils.go
```diff
@@ -408,6 +408,14 @@ func IsNodeFloatValue(node *yaml.Node) bool {
 	return node.Tag == "!!float"
 }
 
+// IsNodeNumberValue will check if a node can be parsed as a float value.
+func IsNodeNumberValue(node *yaml.Node) bool {
```
daveshanley, line 412: This will need a test. Otherwise, the coverage will drop too much.

daveshanley (approved):
LGTM

Issue #117, opened 2023-05-17
Convert `[]error` to use multi errors available in 1.20


thrawn01:
Thanks for the wonderful library! After converting from a different library, I have a single suggestion!

The current code base returns multiple errors which is inconvenient to use if multiple errors are not useful to the caller. We could instead create an error that implements the `Unwrap() []error` method which can be used to retrieve multiple errors.

```go
// Existing signature
RenderAndReload() ([]byte, Document, *DocumentModel[v3high.Document], []error)
// Would become
RenderAndReload() ([]byte, Document, *DocumentModel[v3high.Document], error)
```

Example implementation
```go
type MultiError struct {
	errs []error
}

func (e *MultiError) Error() string {
	var b []byte
	for i, err := range e.errs {
		if i > 0 {
			b = append(b, '\n')
		}
		b = append(b, err.Error()...)
	}
	return string(b)
}

func (e *MultiError) Unwrap() []error {
	return e.errs
}
```

The conversion would be easy, but would break the compatibility. As such it should probably wait for a v1 or v2 release. 

See https://pkg.go.dev/errors#Join

n0rig:
@daveshanley is this still the approach you want to be followed? If so, I can work on this.

jxsl13:
I'd suggest just using errors.Join and return a single error.

daveshanley:
I would avoid working on this until the library is ready for 1.0, it's a major breaking change to the top level of the library in terms of signature. It's a small change with big consequences.

stevenh:
@daveshanley I would suggest doing it ASAP as that will result in less people having to alter their code.

PR #123, opened 2023-05-30
feat: Public methods now return MultiError instead of []error

thrawn01:
## Purpose
The current code base returns multiple errors which is inconvenient to use if multiple errors are not useful to the caller. Instead this PR creates an error that implements the `Unwrap() []error` method which can be used to retrieve multiple errors. This interface was introduced in `go 1.20`

## Implementation
* Added `type MultiError struct` 
* Added `errorMsg()` and `errorMsgf()` (Although that could be narrowed down to just `errorMsg()` with the capability to handle optional formatted text)

## Versioning
This PR significantly breaks compatibility of the public interfaces. If this PR is acceptable, we should bump to `v1` before merging.

## Postscript
My editor naturally indents with a TAB, and as this project uses SPACES for indentation this PR mixes TABS and SPACES. If this PR is to be accepted I'll fix accordingly. 

Reference: https://github.com/pb33f/libopenapi/issues/117

### DIFF
--- a/document.go
+++ b/document.go
@@ -14,7 +14,6 @@
 package libopenapi
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/pb33f/libopenapi/index"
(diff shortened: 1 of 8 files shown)


daveshanley:
Thank you very much for doing this work. I know the change is small, but as you correctly determined, it's a large breaking change for existing consumers - however, it's the right change to make.

We will keep this PR warm until we're ready for 1.0, it should stay healthy and rebase easily when it comes time.

(The build is failing btw :-)

n0rig:
@thrawn01 do you have time to address the build issues and conflicts?

thrawn01:
I unfortunately do not, not until the end of the year  Feel free to close or take the code as is if you want to use this code as the basis for a new PR.

No need to credit me.

PR #129, opened 2023-06-17
Use EqualFold ti avoid memory reallocations

bitomaxsp:
On very large specs `ToLower` becomes a bottleneck since it reallocates a lot


bitomaxsp:
First profile with ToLower
profile001

Second with EqualFold
profile004

daveshanley (approved):
Oh wow, this is great! Thank you!

daveshanley:
I will include this with the `v0.9.0` #128 release

PR #136, opened 2023-06-22
Fix race condition in Schema & Comparisons 

thrawn01:
### Purpose
Found race conditions in `comparison_functions.go` and `high/base/schema.go`. 

### Implementation
* Removed asynchronous nature of `buildProps()` in `NewSchema()`
* In `CheckMapForChangesWithComp()` the lock now covers writes `p[k].GetValueNode().Value = k` when they occur.

### DIFF
--- a/datamodel/high/base/schema.go
+++ b/datamodel/high/base/schema.go
@@ -431,15 +431,13 @@ func NewSchema(schema *base.Schema) *Schema {
 	}
 
 	completeChildren := 0
-	completedProps := 0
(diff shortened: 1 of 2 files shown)


thrawn01:
I do not know why the build is failing, that test and the test suite passes on my local box

```
$ go version
go version go1.20.4 darwin/arm64
```

I'm unable to restart the build to understand if this is a flapping test or some other issue.

daveshanley:
I am not sure why the build failed. Probably DNS somehow.

It's always DNS ;)

daveshanley (approved):
LGTM

Issue #119, opened 2023-05-19
Provide datastructures that retain the original order of items in the document


TristanSpeakEasy:
Ideally we would be able to iterate through a document in the order of items (think paths, operations etc, etc) but due to the use of go native maps this isn't possible with the high level API.

I would recommended switching to using https://github.com/wk8/go-ordered-map or similar to provide this functionality which might also help with this other issue I opened https://github.com/pb33f/libopenapi/issues/109 to ensure rendering new items out can retain an order.

daveshanley:
This is a good idea; it's a heavy lift, however. A good `v0.2.0` feature I think.

diamondburned:
Seeing how we want to preserve API backwards compatibility, is there a chance that such a feature would be merged in, or would we have to wait until at least v0.2?

diamondburned:
I was [half way through implementing this](https://github.com/diamondburned/libopenapi/tree/giveup) when I noticed that I don't really need any of this to preserve object property order. All I needed to do was use the `GoLow()` method and yoink the line number info:

```go
propertyLines := make(map[string]int, len(propertyNames))
for _, name := range propertyNames {
    lowProperty := schema.GoLow().FindProperty(name)
    propertyLines[name] = lowProperty.ValueNode.Line
}
sort.Slice(propertyNames, func(i, j int) bool {
    return propertyLines[propertyNames[i]] < propertyLines[propertyNames[j]]
})
```

daveshanley:
We have this feature being cooked up by in #138 and #148.

TristanSpeakEasy:
We solved this in the interim with:

```
func SortMap[hV any, lV any](highMap map[string]hV, lowMap map[low.KeyReference[string]]lV, maintainOriginalOrder bool) *orderedmap.OrderedMap[string, hV] {
	if highMap == nil {
		return nil
	}

	if !maintainOriginalOrder || len(lowMap) == 0 {
		return utils.MapToOrderedMap(highMap)
	}

	type val struct {
		key low.KeyReference[string]
		val hV
	}

	var vals []val

	for k := range lowMap {
		vals = append(vals, val{key: k, val: highMap[k.Value]})
	}

	sort.Slice(vals, func(i, j int) bool {
		return vals[i].key.GetKeyNode().Line < vals[j].key.GetKeyNode().Line
	})

	om := orderedmap.New[string, hV]()
	for _, v := range vals {
		om.Set(v.key.Value, v.val)
	}

	return om
}
```

and

[...]

PR #138, opened 2023-07-12
WIP: Consistent ordering of documentation

Baliedge:
https://github.com/pb33f/libopenapi/issues/119

Preserve order of paths as they were added to the document.
Utilize `github.com/wk8/go-ordered-map` instead of `map`.
* BREAKING CHANGE: This creates a breaking change around usage of `Paths` and `Schema` structures.  Use `orderedmap.OrderedMap` interface (`Get`/`Set`/`Len` methods) instead of `map` built-in.

Depends on:
* https://github.com/pb33f/libopenapi/pull/148


daveshanley (commented):
Looking good so far, this is the correct way to build the feature. The only comments I really have are why the `MustGet` signature? Why not just `Get`? What am I missing?

daveshanley:
Are we any closer to having this ready to go in?

daveshanley:
@Baliedge Are you able to complete this work or will we need to take over the PR to pull it over the finish line?

TristanSpeakEasy:
This has become critical to some upcoming work we have so I am going to see what it will take to pick up and get over the line. So @daveshanley or @Baliedge if you have any pointers on some keys areas that still need some work let me know otherwise I will move forward trying to address some of the "goals" we need to achieve with ordered maps, mainly:

- Support iterating through the high model in the same order as the OpenAPI document
- Render documents out deterministically
- Render documents out matching the original order (if just making light modifications to the document)
- Control the order of the rendered out docs, for example when the order of path items or components are specifically reordered
- Any maps stored within `example` or `examples` fields in schemas can maintain order both on read/write

Baliedge:
@daveshanley @TristanSpeakEasy After some time using this code in our production codebase, I have some apprehension about the design used here and some design details could use consideration.  Most of your requirements are met with this PR with exceptions.  Mainly, I have these issues in mind:

- Weakly typed components, such as `examples`, don't properly use ordered maps.  Once an `orderdmap.Map` is marshaled to YAML, it unmarshals to `map` and breaks determinism.
  - While annoying, it hasn't been considered a showstopper issue for us.
- Refactoring existing data models defined in code from `map` to `orderedmap.Map` is very tedious.  Basic read/write access of `orderedmap.Map` objects is more tedious.  Is there a better design than enforcing a custom map type?
  - I see that the `low` models reference line numbers in each node, which helps keep determinism if used properly.
  - Perhaps the proper solution is a redesigned `high` model to eliminate use of `map`s entirely in favor of arrays (or ordered map).
  - Either way, this is a very heavy code change.

At the moment, I'm not comfortable asking for approval on this PR until these issues have been considered.

TristanSpeakEasy:
Thanks @Baliedge some good points there.

> Weakly typed components, such as examples, don't properly use ordered maps. Once an orderdmap.Map is marshaled to YAML, it unmarshals to map and breaks determinism.
> While annoying, it hasn't been considered a showstopper issue for us.

For us this is actually one of the bigger show stoppers and on the "read" side we have a lot of nasty code to workaround this and on the write side we have no solution apart from post processing the yaml and sorting alphabetically but our customer's are annoyed at this as its breaking their intentions.

We have been thinking about an API for examples that basically makes it a union of the different "types" an example can be (ie object/map, array, primitive value) instead of just treating it as "any"

> Basic read/write access of orderedmap.Map objects is more tedious.

Agreed the interface is way less idiomatic but we have basically eaten this cost at the moment and have a lot of code all over the place that will take the high maps and convert it to an ordered map using the low line numbers of the key nodes to control insertion order into the ordered map.

So we already access everything via `.Get()` or iterate through the map using pairs and just accepted this as the cost of dealing with ordered data.

I'm not so sure treating this would be any less tedious as there is actually a lot of usage patterns where you do want to do key value lookups.

-----------------------------------------------------------------------

Personally for us we are happy enough with a switch to ordered maps in the high API as it larger removes some many gotchas when trying to deal with "existing" documents and we are building lots of tools that are trying to add value to the documents but in doing so at the moment we tend to have a lot of churn and diffs between the original doc and the modified one are just unwieldy and largely the features we are building are avoided because of the high churn in document structure.

I have just today been merging in the latest main to your branch and fixing some of the failing tests as if this PR landed for us it would be a measurable improvement both to our own code quality but would meet the expectations of our customers much more

Baliedge:
@TristanSpeakEasy Thanks for the quick feedback.  I'm agreement with everything you've said. 
 Until now it was uncertain how acceptable all these breaking changes would be to the APIs due to the data type changes.

With all this understood, the only concerns I might have are edge cases where `maps` were neglected in the API and would result in additional future breaking changes once found.  Additional attention to unit testing could prevent this, but I'm at the disadvantage of not being the expert on OpenAPI spec and every API feature available in this tool.  Some oversight from your team on this would be beneficial.

Aside from that, I am done with my feature changes.

Baliedge:
@TristanSpeakEasy Sure, fork it and do what you need.

daveshanley:
There is no real way around the need to break the high level APIs in order to implement this feature. The design of the `NodeRenderer` uses the underlying low model to determine the order in which to render, so it will keep as much as it can in order, where there is data available. 

Once we have stable ordering / rendering coming from the high level model, this matures this feature and solves a lot of future functionality for the library.

TristanSpeakEasy:
@Baliedge shall we close this PR?

PR #140, opened 2023-07-17
chore: add some more tests for yaml anchor support

ThomasRooney:


datamodel/low/v3/create_document_test.go
```diff
@@ -669,6 +669,59 @@ externalDocs:
...
+	assert.Equal(t, getParams, postParams)
+
+	// check post request body
+	responses := examplePath.GetValue().Get.GetValue().GetResponses().Value.(*Responses)
+	assert.NotNil(t, responses)
+	jsonGet := responses.FindResponseByCode("200").GetValue().FindContent("application/json")
+	assert.NotNil(t, jsonGet)
+
+	// Should this work? It doesn't
+	//postJsonType := examplePath.GetValue().Post.GetValue().RequestBody.GetValue().FindContent("application/json")
```
daveshanley, line 721: It should work. Which means this is a gap.

PR #142, opened 2023-07-18
path unescape url when resolving remote reference

### LINKED ISSUE #141: Remote references: some urls may have encoded characters.
Assuming an open api yaml document contain a remote reference with the following url: `https://some-url.com/open-api.yml#/paths/~1object~1{objectId}~1test/post/responses/200`, for example:
```
      responses:
        200:
          $ref: "https://some-url.com/open-api.yml#/paths/~1object~1{objectId}~1test/post/responses/200"
```

Technically, the url is correct and can be used by [libopenapi](https://github.com/pb33f/libopenapi). However, open api [documentation](https://swagger.io/docs/specification/using-ref/) specifies that the $ref urls should be [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) compliant. This means that any special characters like `{` or `(` should be percent encoded. More details in sections [2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4) and  [2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1). 

For example, if the above url is used in swagger editor, the following error is displayed:

editor swagger io_ (2)

URL encoding the `{` and the `}` characters in the url fixes the issue: `https://some-url.com/open-api.yml#/paths/~1object~1%7BobjectId%7D~1test/post/responses/200`. 

However this breaks libopenapi remote reference resolution which makes the document processing end with errors. 

Solution: performing a path unescape on urls before resolving remote references. This should be relatively safe and would presumably be backward compatible. Additionally, it would make libopenapi comply with RFC3986.

zak905:
closes #141

### DIFF
--- a/index/find_component.go
+++ b/index/find_component.go
@@ -189,6 +189,11 @@ func (index *SpecIndex) lookupRemoteReference(ref string) (*yaml.Node, *yaml.Nod
     } else {
         query = "$"
     }
+    
+    query, err := url.PathUnescape(query)
+    if err != nil {
+        return nil, nil, err
+    }
 
     // remove any URL encoding
     query = strings.Replace(query, "~1", "./", 1)
@@ -282,6 +287,11 @@ func (index *SpecIndex) lookupFileReference(ref string) (*yaml.Node, *yaml.Node,
         query = fmt.Sprintf("$%s", strings.ReplaceAll(uri[1], "/", "."))
     } else {
         query = "$"
+    }
+
+    query, err := url.PathUnescape(query)
+    if err != nil {
+        return nil, nil, err
     }
 
     // remove any URL encoding
--- a/index/find_component_test.go
+++ b/index/find_component_test.go
@@ -181,6 +181,50 @@ paths:
 	assert.Equal(t, "form", crsParam.Node.Content[9].Value)
 }
 
+func TestSpecIndex_LocateRemoteDocsWithMalformedEscapedCharacters(t *testing.T) {
+	spec := `openapi: 3.0.2
+info:
+  title: Test
+  version: 1.0.0
+paths:
+  /test:
+    get:
+      parameters:
+        - $ref: "https://petstore3.swagger.io/api/v3/openapi.yaml#/paths/~1pet~1%$petId%7D/get/parameters"`
+
+	var rootNode yaml.Node
+	_ = yaml.Unmarshal([]byte(spec), &rootNode)
+
+	c := CreateOpenAPIIndexConfig()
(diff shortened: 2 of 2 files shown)


index/find_component.go
```diff
@@ -189,6 +189,11 @@ func (index *SpecIndex) lookupRemoteReference(ref string) (*yaml.Node, *yaml.Nod
     } else {
         query = "$"
     }
+    
+    query, err := url.PathUnescape(query)
```
daveshanley, line 193: Nice upgrade!

My only comment is that this branch will drop code coverage without a test to check for the invalid case.

index/find_component.go
```diff
@@ -284,6 +289,11 @@ func (index *SpecIndex) lookupFileReference(ref string) (*yaml.Node, *yaml.Node,
         query = "$"
     }
 
+    query, err := url.PathUnescape(query)
+    if err != nil {
```
daveshanley, line 293: Same comment as above about the coverage.

daveshanley (changes requested):
Thank you, this is a nice upgrade. 

Please add a couple of small tests to ensure the negative is handled correctly.

index/find_component.go
zak905: thank you. Sure I am going to add some testing

index/find_component.go
zak905: allright

daveshanley:
Did you decide to stop working on this?

zak905:
Hi, Sorry something went wrong when I force pushed. Github automatically closed this when it saw that the main branch has no additional commits compared to the upstream main branch. Working on it.

zak905:
@daveshanley FYI, I have added some units.

zak905:
Thanks, it seems like codecove server is encoutring some issues:
```
The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.<
````

Would you mind relaunching the build please ?

daveshanley:
I re-ran the build, there is just one single line missing from the coverage (that's causing it to be red)

https://app.codecov.io/gh/pb33f/libopenapi/pull/142/blob/index/find_component.go#L294

zak905:
thanks. I added an additional test to cover the missing line

zak905:
thanks @daveshanley

zak905:
@daveshanley, are planning on creating a tag/release for this patch ?

daveshanley:
It was released in `v0.9.7`

zak905:
great thanks!

PR #161, opened 2023-08-25
Expose key node in low schema proxy

bitomaxsp:
Hi Dave,
We came to a point where we need to get access to a key node for the objects. As I figured the key node is present in the low proxy, but there are no methods to access it. When I've added a method, i noticed that key node (kn) is always empty.

I made an attempt to fix it. Please let me know what do you think?  Or is there any better way to get access to a key node.

Main changes are:
```go
// On Buildable key was added
Build(key, value *yaml.Node, idx *index.SpecIndex) error
```

```go
// On low proxy

// GetKeyNode will return the yaml.Node pointer that is a key for value node.
func (sp *SchemaProxy) GetKeyNode() *yaml.Node {
	return sp.kn
}
```

### DIFF
--- a/datamodel/low/base/schema_proxy.go
+++ b/datamodel/low/base/schema_proxy.go
@@ -54,10 +54,12 @@ type SchemaProxy struct {
 }
 
 // Build will prepare the SchemaProxy for rendering, it does not build the Schema, only sets up internal state.
-func (sp *SchemaProxy) Build(root *yaml.Node, idx *index.SpecIndex) error {
-	sp.vn = root
+// Key maybe nil if absent.
+func (sp *SchemaProxy) Build(key, value *yaml.Node, idx *index.SpecIndex) error {
+	sp.kn = key
+	sp.vn = value
 	sp.idx = idx
-	if rf, _, r := utils.IsNodeRefValue(root); rf {
+	if rf, _, r := utils.IsNodeRefValue(value); rf {
 		sp.isReference = true
 		sp.referenceLookup = r
 	}
@@ -125,6 +127,11 @@ func (sp *SchemaProxy) SetReference(ref string) {
 // IsSchemaReference()
 func (sp *SchemaProxy) GetSchemaReference() string {
 	return sp.referenceLookup
+}
+
+// GetKeyNode will return the yaml.Node pointer that is a key for value node.
+func (sp *SchemaProxy) GetKeyNode() *yaml.Node {
+	return sp.kn
 }
 
 // GetValueNode will return the yaml.Node pointer used by the proxy to generate the Schema.
--- a/datamodel/low/extraction_functions.go
+++ b/datamodel/low/extraction_functions.go
@@ -118,7 +118,7 @@ func LocateRefNode(root *yaml.Node, idx *index.SpecIndex) (*yaml.Node, error) {
 
 // ExtractObjectRaw will extract a typed Buildable[N] object from a root yaml.Node. The 'raw' aspect is
 // that there is no NodeReference wrapper around the result returned, just the raw object.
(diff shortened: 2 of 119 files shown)


daveshanley:
I'll review it ASAP, for other readers - would you be able to put a small blurb of example code of how to use the code?

datamodel/low/base/schema_proxy.go
```diff
@@ -127,6 +129,11 @@ func (sp *SchemaProxy) GetSchemaReference() string {
 	return sp.referenceLookup
 }
 
+// GetKeyNode will return the yaml.Node pointer that is a key for value node.
```
daveshanley, line 132: I don't know how I missed this.

datamodel/low/base/schema_proxy.go
```diff
@@ -54,10 +54,12 @@ type SchemaProxy struct {
 }
 
 // Build will prepare the SchemaProxy for rendering, it does not build the Schema, only sets up internal state.
-func (sp *SchemaProxy) Build(root *yaml.Node, idx *index.SpecIndex) error {
-	sp.vn = root
+// Key maybe nil if absent.
+func (sp *SchemaProxy) Build(key, value *yaml.Node, idx *index.SpecIndex) error {
```
daveshanley, line 58: This looks fine, it's a breaking change however, so we will need to bump to `v0.10`

datamodel/low/extraction_functions.go
```diff
@@ -118,7 +118,7 @@ func LocateRefNode(root *yaml.Node, idx *index.SpecIndex) (*yaml.Node, error) {
 
 // ExtractObjectRaw will extract a typed Buildable[N] object from a root yaml.Node. The 'raw' aspect is
 // that there is no NodeReference wrapper around the result returned, just the raw object.
-func ExtractObjectRaw[T Buildable[N], N any](root *yaml.Node, idx *index.SpecIndex) (T, error, bool, string) {
+func ExtractObjectRaw[T Buildable[N], N any](key, root *yaml.Node, idx *index.SpecIndex) (T, error, bool, string) {
```
daveshanley, line 121: Will require minor version bump.

daveshanley:
It looks OK, but it does break signatures for some public methods, which means this has to be a minor version bump. @TristanSpeakEasy, would you like to weigh in?

TristanSpeakEasy:
> It looks OK, but it does break signatures for some public methods, which means this has to be a minor version bump. @TristanSpeakEasy, would you like to weigh in?

We aren't using this particular API so doesn't impact us, but just generally we are fine with breaking changes and will generally put in the work on our side to keep us up to date with any new updates and changes to libopenapi

datamodel/low/base/schema_proxy.go
bitomaxsp: It is indeed breaking change, though i'm not sure how many out there are using Build API. I know we do, but for us it's no issue to change.

datamodel/low/extraction_functions.go
bitomaxsp: yup

bitomaxsp:
> I'll review it ASAP, for other readers - would you be able to put a small blurb of example code of how to use the code?

Where should I put an example code ?

daveshanley:
> > I'll review it ASAP, for other readers - would you be able to put a small blurb of example code of how to use the code?
> 
> Where should I put an example code ?

Actually don't worry about it, it's such a low-level API, I doubt there are many other folks that will need to understand the change.

daveshanley (approved):
LGTM, it's a breaking change - but a very low-level one, and I am doubtful it will have any impact.

daveshanley:
Available in `v0.10.0`

PR #165, opened 2023-09-08
fix: add missing const value to schema

TristanSpeakEasy:
Libopenapi was missing the ability to access the [const field](https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values) of a schema.

This adds that

### DIFF
--- a/datamodel/low/base/schema_test.go
+++ b/datamodel/low/base/schema_test.go
@@ -405,7 +405,8 @@ contentEncoding: fish64
 contentMediaType: fish/paste
 items: true
 examples:
-  - testing`
+  - testing
+const: tasty`
 
 	var rootNode yaml.Node
 	mErr := yaml.Unmarshal([]byte(testSpec), &rootNode)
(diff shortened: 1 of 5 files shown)


datamodel/low/base/schema_test.go
```diff
@@ -415,23 +415,23 @@ examples:
...
+	assert.False(t, sch.ExclusiveMinimum.Value.IsA())
+	assert.True(t, sch.ExclusiveMaximum.Value.IsB())
+	assert.Equal(t, float64(12), sch.ExclusiveMinimum.Value.B)
+	assert.Equal(t, float64(13), sch.ExclusiveMaximum.Value.B)
+	assert.Len(t, sch.Examples.Value, 1)
+	assert.Equal(t, "testing", sch.Examples.Value[0].Value)
+	assert.Equal(t, "fish64", sch.ContentEncoding.Value)
+	assert.Equal(t, "fish/paste", sch.ContentMediaType.Value)
+	assert.True(t, sch.Items.Value.IsB())
+	assert.True(t, sch.Items.Value.B)
```
daveshanley, line 434: Would you be able to add an assertion in here to cover the const?

TristanSpeakEasy: done

daveshanley: LGTM!

PR #166, opened 2023-09-13
Added new `Render` method to top level document.

daveshanley:
Like `RenderAndReload` this method will render / print out the document, however it won’t read it back in and reload the model from the ground up. This is a non-destructive method that does not reload the model from the rendered document.

If you don’t use the low model or you don’t care that the high level model and low-level models are out of sync? No problem. Otherwise keep this in mind that the low level model will only represent the original un-mutated document, as will the index.

Also ran go fmt, cleaned up some indents.


daveshanley:
The drop in patch coverage in `find_component` is a ghost, there were no changes there and logic was added, not removed. ignoring.

Issue #163, opened 2023-09-01
Race conditions when used as middleware...


jasonquigleyhpecom:
Hi Dave,

I am trying to use `libopenapi` in `gin` middleware. I initialise a document model with `BuildV3Model()` and then return a `gin.HandlerFunc`. All pretty standard stuff. 

When I run a lot of tests in parallel I get race errors. If I initialise a model with every request, no more race errors. This would hint that the model is being mutated somewhere. What is your gut feeling on this?

Thanks,
Jason.

daveshanley:
Are you just reading? no writing? what race errors are you getting? It's hard to tell without knowing where the model gets upset and how you're using it.

jasonquigleyhpecom:
The model is just being used for validation, no writing.

I'll clean up a small repo so you can see what I see.

jasonquigleyhpecom:
Here's the trace:

```
WARNING: DATA RACE
Write at 0x00c000fa7238 by goroutine 162:
  github.com/pb33f/libopenapi/datamodel/high/base.(*DynamicValue[...]).MarshalYAMLInline()
      /home/jquigley/projects/libopenapi/datamodel/high/base/dynamic_value.go:98 +0x4a
  github.com/pb33f/libopenapi/datamodel/high/base.(*DynamicValue[...]).MarshalYAMLInline()
      /home/jquigley/projects/libopenapi/datamodel/high/base/dynamic_value.go:97 +0x29
  github.com/pb33f/libopenapi/datamodel/high.(*NodeBuilder).AddYAMLNode()
      /home/jquigley/projects/libopenapi/datamodel/high/node_builder.go:536 +0x1b36
  github.com/pb33f/libopenapi/datamodel/high.(*NodeBuilder).Render()
      /home/jquigley/projects/libopenapi/datamodel/high/node_builder.go:274 +0x6da
  github.com/pb33f/libopenapi/datamodel/high/base.(*Schema).MarshalYAMLInline()
      /home/jquigley/projects/libopenapi/datamodel/high/base/schema.go:490 +0x7c
  github.com/pb33f/libopenapi/datamodel/high/base.(*Schema).RenderInline()
      /home/jquigley/projects/libopenapi/datamodel/high/base/schema.go:477 +0x2e
  github.com/pb33f/libopenapi-validator/schema_validation.validateSchema()
      /home/jquigley/projects/libopenapi-validator/schema_validation/validate_schema.go:76 +0xe4
[...]

PR #172, opened 2023-09-18
Added `SchemaProxy` mutex #163

daveshanley:
Rendering causes bits to be set on structs in the model. Concurrency causes race issues with reading a writing of these bits, as there is no locking in the model.

Until now. locking prevents concurrent renders that use a shared model from conflicting with one another.

Addresses #163 and https://github.com/pb33f/libopenapi-validator/issues/23

### DIFF
--- a/datamodel/high/base/schema.go
+++ b/datamodel/high/base/schema.go
@@ -176,50 +176,50 @@ func NewSchema(schema *base.Schema) *Schema {
 		s.UniqueItems = &schema.UniqueItems.Value
 	}
 	if !schema.Contains.IsEmpty() {
-		s.Contains = &SchemaProxy{schema: &lowmodel.NodeReference[*base.SchemaProxy]{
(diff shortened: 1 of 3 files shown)


jxsl13:
why is the mutex a pointer?

jxsl13:
should a pointer to a mutex be copied?
I think that might lead to problems.
Go has a builtin detection for when a struct is copied that contains a mutex. If I'm not mistaken, a pointer mutex is not caught by that detection.

daveshanley:
I don't know, I haven't seen any literature saying that it's any different from a struct.

https://go.dev/src/sync/mutex.go

The receiver on `Mutex` is a pointer for `Lock` and `Unlock`

jxsl13:
I am not sure on that one either, just having seen warnings pop up in the terminal when a struct with a mutex was copied.

daveshanley:
I've not seen any warnings pop up and the behavior is as I expect, tests pass accordingly. If any strangeness appears with this functionality, then I'll drop it back to a struct :)

Issue #130, opened 2023-06-19
Incorrectly Failing Validation for Circular Reference (Array)


ndimares:
@daveshanley I'm running into a circular reference issue that I think is actually valid syntax:
```
ProductCategory:
      type: "object"
      properties:
        id:
          type: "string"
          description: ""
        name:
          type: "string"
          description: "The name of the category"
        children:
          type: "array"
          items:
            $ref: "#/components/schemas/ProductCategory"
          description: "Array of sub-categories in the same format."
      required:
        - "id"
        - "name"
        - "children"
```
I think this gets flagged as an infinite circular reference because it is a recursive structure that has a **required** field. However, I feel like it should be valid, because an empty array seems like a valid way to break the circle.

ndimares:
Note, this is very similar to: https://github.com/pb33f/libopenapi/issues/113

daveshanley:
I have been thinking about this. If I am a compiler, this looks 100% cyclical to me. That's the point of the warning. Depending on how the spec is used, it may or may not create problems down the line.

What should I tell the parser here? If it **MUST** follow the reference to ensure validity (required) and the chain immediately loops back on itself - there is no way to know when to stop.

The flag is there for you to know that this is a problem. However, it's up to you to decide whether you care. The library will continue operating normally (that's why the `SchemaProxy` design exists, to stop run-away chaos.

TristanSpeakEasy:
Hey @daveshanley we are struggling to deal with this one, while we could just ignore the error and continue with the model I don't feel we get enough info back from libopenapi to be able to make the correct decision on a circular reference error created by this case. I was largely heading down a road of reimplementing a lot of the recursive walking code in libopenapi to be able to get enough context on things like this nodes parents etc and tracking recursion.

I would largely also consider it equivalent to the special case you have in for not treating non-required properties as circular reference errors, in that you ensure the parse doesn't infinitely recurse on them with some short circuiting and this case could be handled in the same way.

As the above is a legitimate way of designing a document and the json that matches it is perfectly valid I would expect there to be no errors returned from parsing this document (or any that have "valid" circular references vs "invalid" ones there are a few of these "valid" special cases, like non-required properties/arrays/oneOf etc)

daveshanley:
I am not sure how else to determine 'what is a circle' without there being some definitive set of rules on which we all agree and then potentially adding a config to those rules, so we can dial them down for a compatibility mode and then turn it up for hard checking.

daveshanley:
[...]

Issue #113, opened 2023-05-05
Incorrectly failing validation for a circular reference?


ndimares:
Hi @daveshanley,
A [Corner case](https://github.com/speakeasy-api/product-resources/issues/16) brought in by a Speakeasy user. They have modified the Apache Airflow API and are generating clients for it. They [want to define a recursive data structure](https://github.com/speakeasy-api/product-resources/issues/16), but it's tripping a [validation from libopenapi ](https://pb33f.io/libopenapi/circular-references/)(I think).

I actually feel like this should be valid? In this example the children data structure is required but it uses an Anyof, so the circular reference shouldn't be infinite?

daveshanley:
I will have a look asap and see what's up. I haven't looked at the spec/issue yet but just off the top of my head, there are two types of error for circular errors, one is a notification that you have a circle, the other is a bit is set to mark a reference as a infinite loop. This is. Recursive structure that is required but loops indefinitely.

Check the infinite loop bit first to see if it's a genuinely broken loop, or it's it's just letting you know a circle has been found, but it's not a required prop, so it can be allowed to fail.

ndimares:
@daveshanley so I think the issue here is that the required prop is being used, but I actually think the circular reference is valid because it's a tree-like structure where either another tree or a value can be contained.  Basically I think there needs to be subrule to your, "no required circular references *unless an anyOf is being used".

daveshanley:
I haven't been able to look into this yet, i'm on vacation and I'm not looking at code for another week or so. 

AnyOf was a design mistake in my opinion, it creates more chaos than it solves. 

Back online in a week or so.

ndimares:
Also, @TristanSpeakEasy pointed out to me that oneOf rather than anyOf should be used in this scenario. However, I did a quick test and confirmed that the same validation error pops up when you use OneOf.

daveshanley:
This is now configurable from `v0.11.0`: https://github.com/pb33f/libopenapi/pull/173

PR #173, opened 2023-09-19
Added support for configurable levels of circular reference checking …

daveshanley:
Circular references mean different things to different consumers. They may or may not be valid use cases. So after much discussion, I've decided to allow engineers to override my opinions on circular references. 

Polymorphic based references, or array based references may or may not be valid circles, an empty array can break the loop for example. This isn't clear in the spec as there can be no way of knowing. The default position of `libopenapi` is to **not** ignore polymorphic (`anyOf`, `oneOf`, `allOf`) loops, or array based loops. 

Now this can be configured, Choose to ignore polymorphic loops, or array loops, or both. 

To configure the resolver directly, there are two new methods available:

`IgnorePolymorphicCircularReferences()` and `IgnoreArrayCircularReferences()`

The same properties exist on the `datamodel.DocumentConfiguration` struct that allows them to be ignored when building a document. For example:

```go
var d = `openapi: 3.1.0
components:
  schemas:
    ProductCategory:
      type: "object"
      properties:
        name:
          type: "string"
        children:
          type: "object"
          anyOf:
            - $ref: "#/components/schemas/ProductCategory"
          description: "Array of sub-categories in the same format."
      required:
        - "name"
        - "children"`

	config := datamodel.NewClosedDocumentConfiguration()
        
        // ignore polymorphic circular references.
	config.IgnorePolymorphicCircularReferences = true

	doc, err := NewDocumentWithConfiguration([]byte(d), config)
	if err != nil {
		panic(err)
	}

	m, errs := doc.BuildV3Model()

	assert.Len(t, errs, 0)
	assert.Len(t, m.Index.GetCircularReferences(), 0)
```

The second update in this release is that the `datamodel.CircularReferenceResult` contains more detail on
what type of circular reference was located. This is valuable to determine if you should programmatically ignore the loop.

```go
type CircularReferenceResult struct {
	Journey             []*Reference
	Start               *Reference
	LoopIndex           int
	LoopPoint           *Reference
	IsArrayResult       bool   // if this result comes from an array loop.
	PolymorphicType     string // which type of polymorphic loop is this? (oneOf, anyOf, allOf)
	IsPolymorphicResult bool   // if this result comes from a polymorphic loop.
	IsInfiniteLoop      bool   // if all the definitions in the reference loop are marked as required, this is an infinite circular reference, thus is not allowed.
}
```

The `IsArrayResult` will be true if the loop has come through an array

The `PolymorphicTypeResult` will be `oneOf`, `anyOf` or `allOf` depending on which type was used and the
`isPolymorphicResult` value will also be true, as before.

The docs have been updated for some more detail if required:

https://pb33f.io/libopenapi/circular-references/#circular-reference-results

- #113 
- #130

### DIFF
--- a/datamodel/document_config.go
+++ b/datamodel/document_config.go
@@ -40,6 +40,18 @@ type DocumentConfiguration struct {
 	// BypassDocumentCheck will bypass the document check. This is disabled by default. This will allow any document to
 	// passed in and used. Only enable this when parsing non openapi documents.
 	BypassDocumentCheck bool
+
+	// IgnorePolymorphicCircularReferences will skip over checking for circular references in polymorphic schemas.
+	// A polymorphic schema is any schema that is composed other schemas using references via `oneOf`, `anyOf` of `allOf`.
+	// This is disabled by default, which means polymorphic circular references will be checked.
+	IgnorePolymorphicCircularReferences bool
+
+	// IgnoreArrayCircularReferences will skip over checking for circular references in arrays. Sometimes a circular
+	// reference is required to describe a data-shape correctly. Often those shapes are valid circles if the
+	// type of the schema implementing the loop is an array. An empty array would technically break the loop.
+	// So if libopenapi is returning circular references for this use case, then this option should be enabled.
+	// this is disabled by default, which means array circular references will be checked.
+	IgnoreArrayCircularReferences bool
 }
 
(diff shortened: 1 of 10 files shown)


TristanSpeakEasy:
LGTM @daveshanley :shipit:

PR #176, opened 2023-09-22
refactor: remove single iteration loop

nickajacks1:
Simplify code by taking the first element instead of performing a single-iteration loop.

Disclaimer: I wasn't actually 100% sure if there was some magic going on that I didn't realize.
Discovered using https://github.com/golangci/golangci-lint

### DIFF
--- a/renderer/schema_renderer.go
+++ b/renderer/schema_renderer.go
@@ -7,14 +7,15 @@ import (
 	cryptoRand "crypto/rand"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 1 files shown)


daveshanley:
This is by design, removing it will result in incorrect results. Thats why there is a comment on the break. I cannot accept this unfortunately, it will break the library.

daveshanley (approved):
Thank you for fixing this logic. I don't know what I was thinking when I wrote it, I must have been distracted.

Issue #175, opened 2023-09-22
Recent fix to indexing enums causes not all enums to be indexed


TristanSpeakEasy:
The change here https://github.com/pb33f/libopenapi/pull/169 caused some of our tests to start failing as all the enums we were expecting to be in the index are no longer in the index.

For example in this spec:

```
openapi: 3.1.0
servers:
  - url: http://localhost:8080
paths:
  /test:
    get:
      operationId: test
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  enumRef:
                    $ref: '#/components/schemas/enum'
                  enum:
                    type: string
                    enum: [big, small]
                    nullable: true
components:
  schemas:
    enum:
      type: [string, null]
      enum: [big, small]
```

The enum in the component named `enum` is not contained within the index anymore

TristanSpeakEasy:
reverting to `v0.10.3` fixes the issue

daveshanley:
I think this may be a case of 'unfix the bug, we need the bug'. Technically the new behavior is correct, because the property named `enum` is not an enum, so it should not be indexed.

However the bug that exists in `v0.10.3` is something you've come to depend on. 

So, I think the solution here, is the same as the circular references approach. The option to include properties named `enum` into the index, or ignore them (default is to ignore, which is correct in my opinion)

But you can flip a bit in the Index Config or via a method to ignore that check and revert back to the pre bugfix behavior. 

Thoughts?

daveshanley:
OK, I will loop back around on this, more intelligence on deciding what is and is not an enum is needed.

TristanSpeakEasy:
fixed

PR #148, opened 2023-07-27
Generalized parallel processing logic

Baliedge:
Convert specialized parallel processing logic to generalized "translate" functions.
This is needed to prepare for significant datamodel changes supporting consistent ordering (#119) as the changes needed will refactor the same areas of code.
In several places, parallel processing has been implemented to improve performance.  However, the logic is specialized and is prone to concurrency flaws without proper test coverage.  Using a set of generalized functions will decrease the chance of introducing bugs when refactoring for consistent ordering.  e.g. Want to be able to change the data type and container iteration logic without breaking channel logic.

PR #180, opened 2023-10-03
fix!: fixed handling of additionalProperties to handle the bool/json-schema nature better

TristanSpeakEasy:
This PR introduces a fix (via way of a breaking change) to allow a schema in an `additionalProperties` field to be accessed correctly. It also tried to standardise the access of bool/schema properties like `additionalProperties` and `unevaluatedProperties`.

This basically fixes the below schema not being returned as a `SchemaProxy`:

```
type: object
additionalProperties:
  allOf:
    - type: string
    - type: array
      items:
        type: string
```

also incidentally I found a bug that caused an infinite hang and fixed that as part of this

### DIFF
--- a/datamodel/high/base/schema.go
+++ b/datamodel/high/base/schema.go
@@ -63,7 +63,7 @@ type Schema struct {
 
 	// in 3.1 UnevaluatedProperties can be a Schema or a boolean
 	// https://github.com/pb33f/libopenapi/issues/118
-	UnevaluatedProperties *DynamicValue[*SchemaProxy, *bool] `json:"unevaluatedProperties,omitempty" yaml:"unevaluatedProperties,omitempty"`
+	UnevaluatedProperties *DynamicValue[*SchemaProxy, bool] `json:"unevaluatedProperties,omitempty" yaml:"unevaluatedProperties,omitempty"`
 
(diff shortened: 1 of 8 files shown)


index/extract_refs.go
```diff
@@ -332,9 +331,12 @@ func (index *SpecIndex) ExtractRefs(node, parent *yaml.Node, seenPath []string,
 					if len(seenPath) > 0 {
 						lastItem := seenPath[len(seenPath)-1]
 						if lastItem == "properties" {
+							seenPath = append(seenPath, n.Value)
```
TristanSpeakEasy, line 334: @daveshanley also contains a fix for enum extraction that sorts the issues we were seeing after the last change

TristanSpeakEasy: fixes https://github.com/pb33f/libopenapi/issues/175

daveshanley:
Thank you for fixing this!

The next major version jump (because this is a breaking change) will need to include #148 as well. I am not sure how much of a conflict that will create if I merge in that first, then this PR.

Does that work for you?

TristanSpeakEasy:
Yeah I can fix any conflicts when that other branch is merged

Issue #54, opened 2022-12-20
bug: security scheme name starting with `X-` not being added to SecuritySchemes map


TristanSpeakEasy:
If a OpenAPI document defines a security scheme name starting with `X-` it is not showing up in `model.Components.SecuritySchemes`

for example:

```yaml
securitySchemes:
    OAuth:
      flows:
        authorizationCode:
          authorizationUrl: "https://example.com/v1/oauth/authorize"
          refreshUrl: "https://example.com/v1/oauth/token"
          scopes:
            account.manage: This scope grants permissions to perform read/edit/delete actions on Account data
            account.view: This scope grants permissions to perform read only actions on Account data
            openid: This scope grants permissions that enable SSO by granting an id token JWT that stores account data. Not used in v1/account endpoints
          tokenUrl: "https://example.com/v1/oauth/token"
      type: oauth2
    X-API-Key:
      in: header
      name: X-API-Key
      type: apiKey
```

`X-API-Key` can't be found

daveshanley:
Another extension glitch I think, Will address ASAP.

daveshanley:
Resolved in `v0.4.8`

rpmccarter:
Hello! Seems like this is fixed for `X-`, but is still broken for `x-`. I've opened #188, which adds test cases for both cases, and I'm hoping someone can help me fix the underlying bug. Thanks!

PR #188, opened 2023-10-25
add tests for security schemes prefixed with `x-` and `X-`

rpmccarter:
Creates test cases for #54 . Test case passes for `X-`, but fails for `x-`. Not confident enough to contribute a fix, so any help fixing these test cases would be greatly appreciated.

Hopefully adding this will prevent regressions in the future!


daveshanley:
Hi, I have looked into this today again as I was looking through the backlog. 

This use case cannot be supported. `X-` is a valid header prefix, but `x-` is not and cannot be.  Any field that begins with `x-` is considered an extension. 

https://spec.openapis.org/oas/latest.html#specificationExtensions

I can't make this work, it would mean that extensions are not supported for `SecurityScheme` objects, which would violate the spec. When in fact using any non-extension field with a prefix of lowercase `x-` is actually a violation of the spec. 

Thanks.

Issue #200, opened 2023-11-18
Feature request: Add support to mock generator for non required fields


petkostas:
I am using libopenapi in a project to render documentation, and I used the mock generator to render the responses of different operations.
While it was working as expected, I encountered a minor annoyance. I crossed a case where the Media type response had no example and used a schema reference.
The mock generator generated a mock response, but I noticed it was omitting a few fields. After a couple of experiments, I realized that this behavior was related to the use of `required` in the schema object. Removing this provided a complete working example.
Is there a way to switch this (maybe through an `mg.OnlyRequired()` switch? That would greatly help, especially when you prefer to render a complete response example for schemas that include a `required` field.
Let me know if I can help. I can open a PR for this.

P.S. Heh `#200` nice 👍🏼 😄

petkostas:
Hello @daveshanley, a different method would be sufficient for me (and also keeps things clean). The expected behavior would be that when calling that method, you get a mock with all the schema fields regardless of the `required` state.

daveshanley:
Added in `v0.13.12`

A new switch method features on the `renderer.SchemaRenderer` struct called `DisableRequiredCheck()` This will render all properties, regardless of the 'required` property values.

Issue #177, opened 2023-09-30
Improvement: Provide a way to determine where a reference lives


TristanSpeakEasy:
Consider this:

- A top level openapi.yaml references an external ref from one file `otherfile.yaml#/components/schemas/reference1` and another `otherfile2.yaml#/components/schemas/reference2`

Those files might look like:

`otherfile.yaml`
```
components:
  schemas:
    reference1:
      type: array
      items:
      	$ref: '#/components/schemas/item'
    item:
      type: object
      properties:
        field:
          type: string
```

`otherfile2.yaml`
```
components:
  schemas:
    reference2:
      type: object
      additionalProperties:
      	$ref: '#/components/schemas/item'
    item:
      type: object
      properties:
        otherfield:
          type: string
```

The problem is when iterating through the schemas we will eventually come across the `item` from `otherfile.yaml` and the `item` from `otherfile2.yaml` and we may try and use `.ParentProxy.GetSchemaReference()` on the schema to get its "reference" this will return `#/components/schemas/item` for both these `item` schemas and these references may be used to determine their identity but with this there is no way to distinguish them as different references.

It would be good to somehow be able to determine where the schema came from so we can "fully resolve" the reference to `otherfile2.yaml#/components/schemas/item` to ensure they are unique

daveshanley:
This has been on my mind for a while. I have a few use-cases in mind that need this type of upgrade. There needs to be an upgrade to the way indexes track which files they were created from, and where in the tree they sit.

TristanSpeakEasy:
Something like `.ParentProxy.GetReferenceFile()` or having an option to always fully resolve the `.ParentProxy.GetReference()` so it always returns `./components.yaml#/components/schema/user` the reference with the file it was resolved from at the front of the reference

daveshanley:
Added in `v0.13.12`

A new method `GetReferenceOrigin()` exists on `SchemaProxy` that returns `*index.NodeOrigin`


```go 
type NodeOrigin struct {
  // Node is the node in question 
  Node *yaml.Node

  // Line is yhe original line of where the node was found in the original file
  Line int

  // Column is the original column of where the node was found in the original file 
  Column int

  // AbsoluteLocation is the absolute path to the reference was extracted from.
  // This can either be an absolute path to a file, or a URL.
  AbsoluteLocation string

  // Index is the index that contains the node that was located in.
  Index *SpecIndex
}
```

Let me know if you have any issues with it.

PR #201, opened 2023-11-22
v0.13.12

daveshanley:
Added the ability to locate where a reference comes from 

- #177
- #200

A new method `GetReferenceOrigin()` exists on `SchemaProxy` that returns `*index.NodeOrigin`

```go 
type NodeOrigin struct {
  // Node is the node in question 
  Node *yaml.Node

  // Line is yhe original line of where the node was found in the original file
  Line int

  // Column is the original column of where the node was found in the original file 
  Column int

  // AbsoluteLocation is the absolute path to the reference was extracted from.
  // This can either be an absolute path to a file, or a URL.
  AbsoluteLocation string

  // Index is the index that contains the node that was located in.
  Index *SpecIndex
}
```

Also a new switch method features on the `renderer.SchemaRenderer` struct called `DisableRequiredCheck()` This will render all properties, regardless of the 'required` property values.

### DIFF
--- a/datamodel/high/base/schema_proxy.go
+++ b/datamodel/high/base/schema_proxy.go
@@ -4,7 +4,6 @@
 package base
 
 import (
-	"fmt"
 	"github.com/pb33f/libopenapi/datamodel/high"
 	"github.com/pb33f/libopenapi/datamodel/low"
 	"github.com/pb33f/libopenapi/datamodel/low/base"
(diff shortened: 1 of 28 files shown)


daveshanley:
The 0.01 issue is grinding my gears a bit, 100% coverage, yet it groans about a phantom missed line occasionally, even though there is a test for it. hmmm.

PR #205, opened 2023-11-27
feat: introduce ordered maps into libopenapi

TristanSpeakEasy:
Picks up from https://github.com/pb33f/libopenapi/pull/138 and tries to address https://github.com/pb33f/libopenapi/issues/119

The main goals will be:

- Support iterating through the high model in the same order as the OpenAPI document
- Render documents out deterministically
- Render documents out matching the original order (if just making light modifications to the document)
- Control the order of the rendered out docs, for example when the order of path items or components are specifically reordered
- Any maps stored within example or examples fields in schemas can maintain order both on read/write

Done:

- [x] Replace all instances of native map usage
- [x] Improve usage of orderedmap to avoid panics
- [x] Validate some of the changes to tests
- [x] Test new items can be added to docs in expected order
- [x] Test existing items in docs can be reordered
- [x] Improve API for dealing with untyped example/examples fields in schemas


datamodel/high/base/dynamic_value_test.go
```diff
@@ -130,11 +127,10 @@ func TestDynamicValue_MarshalYAMLInline(t *testing.T) {
 
 	// convert node into yaml
 	bits, _ := yaml.Marshal(rend)
-	assert.Equal(t, "properties:\n    rice:\n        type: array\n        items:\n            type: string", strings.TrimSpace(string(bits)))
+	assert.Equal(t, "properties:\n    rice:\n        $ref: '#/components/schemas/rice'", strings.TrimSpace(string(bits)))
```
TristanSpeakEasy, line 130: @daveshanley this changes was expected and points to a difference in behavior being introduced somewhere, but I just wanted to double check what the expected result of this test should be, should it inline the component like the original value? Or is the new value correct and a bug was fixed in the process?

TristanSpeakEasy: okay looks like there was some special casing around maps that is not happening anymore as the ordered maps reflection wise aren't being detected as maps I will need to dig in and see what it would take to match that reflection logic with the ordered maps

daveshanley: All the logic is handled by the `NodeBuilder` in the high level model package. It's what performs the reflection and decides how to handle the data. If there is a low level node available from the high level one (as in, the node was originally pulled in from a spec) then, it uses that low level node as a way to determine order. 

For new content (modified high level model), there is no low level node, so there is no way to determine the order (which is why we have our current status quo)

The `NodeBuilder` will need to learn about the ordered maps, so it can use them effectively and render out correctly. 

Some of that code in there is a bit gnarly, lots of short variable names

TristanSpeakEasy: Okay so for this specific test looks like the previous assertion was actually asserting a bug. This test is calling MarshalInline and therefore all the references should be inline which now happens correctly

TristanSpeakEasy:
> But there are tests failing around the index (some are likely just difference in contents and so byte sizes have changed but I wasn't comfortable just changing the values)

Actually I just fixed these after some other changes the differences were minimal and I was able to validate what was needed.

So @daveshanley its just the failing speakeasy iteration test I need help with

daveshanley:
> > But there are tests failing around the index (some are likely just difference in contents and so byte sizes have changed but I wasn't comfortable just changing the values)
> 
> Actually I just fixed these after some other changes the differences were minimal and I was able to validate what was needed.
> 
> So @daveshanley its just the failing speakeasy iteration test I need help with

Cool, I will check out the branch and take a look ASAP.

datamodel/low/base/schema.go
```diff
@@ -1036,64 +1000,45 @@ func (s *Schema) Build(ctx context.Context, root *yaml.Node, idx *index.SpecInde
...
+		propertyMap := orderedmap.New[low.KeyReference[string], low.ValueReference[*SchemaProxy]]()
 		var currentProp *yaml.Node
-		totalProps := 0
 		for i, prop := range propsNode.Content {
 			if i%2 == 0 {
 				currentProp = prop
 				continue
 			}
 
-			foundCtx := ctx
```
daveshanley, line 1069: Here is the problem, this `foundCtx` is the key to tracking deeply nested schema property builds.  The method `LocateRefNodeWithContext` returns a constructed context that allows the next recursive iteration to track where it is correctly. 

uncomment this line.

datamodel/low/base/schema.go
```diff
@@ -1036,64 +1000,45 @@ func (s *Schema) Build(ctx context.Context, root *yaml.Node, idx *index.SpecInde
...
+			var refNode *yaml.Node
 			if h, _, l := utils.IsNodeRefValue(prop); h {
-				ref, _, _, fctx := low.LocateRefNodeWithContext(ctx, prop, idx)
+				ref, _, _, _ := low.LocateRefNodeWithContext(ctx, prop, idx)
 				if ref != nil {
-					isRef = true
+					refNode = prop
 					prop = ref
 					refString = l
-					foundCtx = fctx
```
daveshanley, line 1079: Add this back in

daveshanley (commented):
This is why the speakeasy tests are failing, this tiny little context is the key. 

The code should look like this: 

```go
            foundCtx := ctx
            if h, _, l := utils.IsNodeRefValue(prop); h {
                if prop.Line == 207 {
                    fmt.Println("here")
                }
                ref, _, _, fctx := low.LocateRefNodeWithContext(ctx, prop, idx)
                if ref != nil {
                    refNode = prop
                    prop = ref
                    refString = l
                    foundCtx = fctx
                } else {
                    return nil, fmt.Errorf("schema properties build failed: cannot find reference %s, line %d, col %d",
                        prop.Content[1].Value, prop.Content[1].Line, prop.Content[1].Column)
                }
            }

            sp := &SchemaProxy{ctx: foundCtx, kn: currentProp, vn: prop, idx: idx}
```

daveshanley:
OK so the problem is simple, you removed the context being extracted with each reference lookup, so the `SchemaProxy` has the wrong context attached to it (it's a parent's context). Which is why it can't find the `enum`. It's looking in the wrong index. 

I added the context back in (see my code above) and the tests pass. 

When it comes to the digital ocean tests failing, it's because they are modifying files every few days, so the rolodex size changes. I am going to strip these size checks out and move them to fixed specs, so these tests don't need to be continually updated. 

Just update the values.. i.e. 

```
expected: 1270079
actual  : 1273069
```

These are byte size values of everything sucked in from the repo. So it's unstable and needs to be moved.

TristanSpeakEasy:
@daveshanley that is amazing thanks, I'll get this all button up and ready to go

TristanSpeakEasy:
@daveshanley I will also have some PRs for `libopenapi-validator` and `vacuum` ready to go once this is merged as part of this work as I needed to upgrade them for my testing

TristanSpeakEasy:
It is looking like this is passing our test suite, (still need to run our regression tests)

But I will cleanup the missing code coverage above tomorrow (or Monday) and then pending any feedback or request for changes, I feel pretty good about the state of this

daveshanley:
> Added a failing test with index resolution that I can't seem to figure out, found this in our test suite but haven't been able to nail down where the regression came in https://github.com/speakeasy-api/libopenapi/blob/ordered-libopenapi/document_test.go#L1175. @daveshanley would appreciate some help tracking down the issue have spent more hours than I care to admit trying to understand what is going on, but it looks like its looking for a reference in the wrong file.
> 
> It returns errors like: `"time":"2023-12-04T10:19:05.104578835Z","level":"ERROR","msg":"unable to locate reference anywhere in the rolodex","reference":"/home/trist/workspace/oss/libopenapi/test_specs/advancecallbackreferences/min-callbacks.yaml#/components/pathItems/test-callback-2"`
> 
> but the ref should actually be `/home/trist/workspace/oss/libopenapi/test_specs/advancecallbackreferences/min-components.yaml#/components/pathItems/test-callback-2` ie its looking in the `min-callbacks.yaml` instead of the `min-components.yaml`

I am traveling to Paris today, but I will attempt to debug this as I move around airports.

datamodel/low/v3/paths.go
```diff
@@ -152,12 +175,11 @@ func (p *Paths) Build(ctx context.Context, _, root *yaml.Node, idx *index.SpecIn
 			path := new(PathItem)
```
daveshanley, line 175: Here is the problem. 

Not sure why, but the parent context is being passed down again.  In the `TranslatePipeline` call, we need to make sure we pull the context when looking up a ref node, and pass it down. Change the code to look like this (sorry I can' t make suggestions on code that has not changed)
Screenshot 2023-12-04 at 12 16 46 PM

daveshanley: the test passes once this change is made.

daveshanley:
Let me know when you're ready for me to review.

TristanSpeakEasy:
It is ready to review now we have had it in production for the last couple
of days and no issues yet

daveshanley:
I am working my way through it, should be done tomorrow.

datamodel/low/v3/create_document.go
```diff
@@ -45,6 +46,7 @@ func createDocument(info *datamodel.SpecInfo, config *datamodel.DocumentConfigur
 	idxConfig.BasePath = config.BasePath
 	idxConfig.Logger = config.Logger
 	rolodex := index.NewRolodex(idxConfig)
+	<-info.GetJSONParsingChannel() // Need to wait for JSON parsing to complete before we can index.
```
daveshanley, line 49: I want to double check this does not slow down vacuum too much, I will pull the branch and perform some a/b testing.

document_test.go
```diff
@@ -198,15 +235,13 @@ func TestDocument_RenderAndReload_ChangeCheck_Stripe(t *testing.T) {
 	tc := compReport.TotalChanges()
 	bc := compReport.TotalBreakingChanges()
 	assert.Equal(t, 0, bc)
-	assert.Equal(t, 820, tc)
+	assert.Equal(t, 819, tc)
```
daveshanley, line 238: Do you know what changed? (pun intended)

document_test.go
```diff
@@ -198,15 +235,13 @@ func TestDocument_RenderAndReload_ChangeCheck_Stripe(t *testing.T) {
...
 
 	// there should be no other changes than the 519 descriptions.
-	assert.Equal(t, 1, len(filtered))
-
+	assert.Equal(t, 0, len(filtered))
```
daveshanley, line 241: what was the difference?

document_test.go
```diff
@@ -240,21 +275,18 @@ func TestDocument_RenderAndReload_ChangeCheck_Asana(t *testing.T) {
 
 	assert.Nil(t, errs)
 	tc := compReport.TotalChanges()
-	assert.Equal(t, 21, tc)
```
daveshanley, line 243: is this accurate?

json/json.go
```diff
@@ -0,0 +1,89 @@
+package json
```
daveshanley, line 1: nice! this is great! Ordered JSON!

orderedmap/builder.go
```diff
@@ -0,0 +1,122 @@
...
+
+	for i := 0; i < len(m.Content); i += 2 {
+		if m.Content[i].Value == key {
+			return m.Content[i]
+		}
+	}
+	return nil
+}
+
+func (o *Map[K, V]) FindValueUntyped(key string) any {
```
daveshanley, line 108: Docs?

orderedmap/builder.go
```diff
@@ -0,0 +1,122 @@
...
+
+type hasValueUntyped interface {
+	GetValueUntyped() any
+}
+
+type findValueUntyped interface {
+	FindValueUntyped(k string) any
+}
+
+func (o *Map[K, V]) ToYamlNode(n NodeBuilder, l any) *yaml.Node {
```
daveshanley, line 40: Docs?

orderedmap/orderedmap.go
```diff
@@ -0,0 +1,215 @@
...
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+
+	wk8orderedmap "github.com/wk8/go-ordered-map/v2"
+)
+
```
daveshanley, line 16: Docs?

orderedmap/orderedmap.go
```diff
@@ -0,0 +1,215 @@
...
+type Pair[K comparable, V any] interface {
+	Key() K
+	KeyPtr() *K
+	Value() V
+	ValuePtr() *V
+	Next() Pair[K, V]
+}
+
+type Map[K comparable, V any] struct {
```
daveshanley, line 25: Docs?

utils/utils.go
```diff
@@ -543,19 +541,6 @@ func ConvertYAMLtoJSON(yamlData []byte) ([]byte, error) {
 	return jsonData, nil
 }
 
-// ConvertYAMLtoJSONPretty will do exactly what you think it will. It will deserialize YAML into serialized JSON.
-// However, this version will apply prefix/indentation to the JSON.
-func ConvertYAMLtoJSONPretty(yamlData []byte, prefix string, indent string) ([]byte, error) {
```
daveshanley, line 548: Why was this removed?

daveshanley (changes requested):
Wow, what a PR! Such a huge lift to turn this around, but provides a real maturity to the library, allowing us to march forward with other higher level features, knowing the lower levels are now solid.

daveshanley:
In regards to coverage, I would really appreciate bringing it up to 99.6 at a minimum. I don't want to lose the `100%` badge we have on the readme, and it rounds up to keep things clean :)

TristanSpeakEasy:
> In regards to coverage, I would really appreciate bringing it up to 99.6 at a minimum. I don't want to lose the `100%` badge we have on the readme, and it rounds up to keep things clean :)

I'll try and achieve this, some of the uncovered lines were a bit hard to determine the best way to test but I'll try a few techniques and then reach out for help if needed

utils/utils.go
TristanSpeakEasy: This was replaced by the Ordered Json here: https://github.com/pb33f/libopenapi/pull/205/files#r1425740867

datamodel/high/v3/path_item.go
```diff
@@ -129,32 +134,65 @@ func (p *PathItem) GoLowUntyped() any {
 	return p.low
 }
 
-func (p *PathItem) GetOperations() map[string]*Operation {
-	o := make(map[string]*Operation)
+func (p *PathItem) GetOperations() *orderedmap.Map[string, *Operation] {
+	o := orderedmap.New[string, *Operation]()
+
+	// TODO: this is a bit of a hack, but it works for now. We might just want to actually pull the data out of the document as a map and split it into the individual operations
```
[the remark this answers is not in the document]
TristanSpeakEasy: So its a hack in terms of we are ordering the disparate operations back into an ordered map as we don't have them available as an ordered map straight from the document. I did try and do the work to pull the ops out as an ordered map and then split into the separate http methods but the work was a bit too low level for me at the time.

datamodel/low/base/schema_proxy.go
```diff
@@ -101,36 +101,6 @@ func (sp *SchemaProxy) GetBuildError() error {
 	return sp.buildError
 }
 
-// IsSchemaReference returns true if the Schema that this SchemaProxy represents, is actually a reference to
-// a Schema contained within Components or Definitions. There is no difference in the mechanism used to resolve the
-// Schema when calling Schema(), however if we want to know if this schema was originally a reference, we won't
-// be able to determine that from the model, without this bit.
-func (sp *SchemaProxy) IsSchemaReference() bool {
```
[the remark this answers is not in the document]
TristanSpeakEasy: I cleaned up how references were done by taking advantage of struct composition with the `Reference` struct which had all these methods and the `IsSchemaReference` methods (as opposed to just `IsReference` weren't doing anything materially different.

Basically this bit of refactoring made it much easier to make sure I was capturing the "refNode" where needed for latter ensuring I was retaining the style for rendering references

datamodel/low/reference.go
```diff
@@ -108,19 +123,14 @@ type NodeReference[T any] struct {
 	// The yaml.Node that is the key, that contains the value.
 	KeyNode *yaml.Node
 
-	// Is this value actually a reference in the original tree?
-	ReferenceNode bool
-
-	// If HasReference is true, then Reference contains the original $ref value.
-	Reference string
```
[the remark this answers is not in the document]
TristanSpeakEasy: https://github.com/pb33f/libopenapi/pull/205#discussion_r1426585177

document_test.go
TristanSpeakEasy: Yeah so a lot of these changes came from the fact I am now more correctly retaining the original styling of the document when it does a round trip through libopenapi, a few things were being picked up as changes just because they were being rendered out differently to how they were originally

document_test.go
TristanSpeakEasy: https://github.com/pb33f/libopenapi/pull/205#discussion_r1426586957

document_test.go
TristanSpeakEasy: https://github.com/pb33f/libopenapi/pull/205#discussion_r1426586957

Just to clarify I believe this is accurate but at the same time a lot of these magic numbers were very hard to validate, what would have been better were more specific tests where you could have been asserting actually changes instead of just counts of changes

datamodel/high/v2/path_item.go
```diff
@@ -101,32 +106,65 @@ func NewPathItem(pathItem *low.PathItem) *PathItem {
...
+		}
+
+		return l.GetKeyNode().Line
+	}
+
+	ops := []op{}
+
 	if p.Get != nil {
-		o[low.GetLabel] = p.Get
+		ops = append(ops, op{name: lowV2.GetLabel, op: p.Get, line: getLine("Get", -7)})
```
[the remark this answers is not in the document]
TristanSpeakEasy: so yeah they just help stabilize the order if we can't find the original line number for the operation.

datamodel/high/node_builder_test.go
```diff
@@ -4,169 +4,147 @@
...
-			g[low.KeyReference[string]{
-				Value:   i,
-				KeyNode: kn,
-			}] = low.ValueReference[any]{
-				ValueNode: &vn,
-				Value:     te.Extensions[i],
-			}
-		}
-
+	Thrig      *orderedmap.Map[string, *plug]                    `yaml:"thrig,omitempty"`
```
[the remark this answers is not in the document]
TristanSpeakEasy: Yeah took a while to fix these tests, I hope they are all testing what was originally tested but it did make it challenging to refactoring as I had to update everything manually

orderedmap/orderedmap.go
TristanSpeakEasy: done

orderedmap/orderedmap.go
TristanSpeakEasy: done

orderedmap/builder.go
TristanSpeakEasy: done

orderedmap/builder.go
TristanSpeakEasy: done

datamodel/high/node_builder.go
```diff
@@ -342,87 +330,8 @@ func (n *NodeBuilder) AddYAMLNode(parent *yaml.Node, entry *NodeEntry) *yaml.Nod
...
-		p := utils.CreateEmptyMapNode()
-
-		// build out each map node in original order.
-		for _, cv := range orderedCollection {
-			n.AddYAMLNode(p, cv)
-		}
-		if len(p.Content) > 0 {
-			valueNode = p
-		}
+		panic("only ordered maps are supported")
```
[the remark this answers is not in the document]
TristanSpeakEasy: done

datamodel/high/node_builder.go
```diff
@@ -177,53 +184,36 @@ func (n *NodeBuilder) add(key string, i int) {
...
-			line := make([]int, l)
-			for q, ky := range value.MapKeys() {
-				if we, wok := ky.Interface().(low.HasKeyNode); wok {
-					line[q] = we.GetKeyNode().Line
-				}
-			}
-			sort.Ints(line)
-			nodeEntry.Line = line[0]
-
+			panic("only ordered maps are supported")
```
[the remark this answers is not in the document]
TristanSpeakEasy: done

TristanSpeakEasy:
> In regards to coverage, I would really appreciate bringing it up to 99.6 at a minimum. I don't want to lose the `100%` badge we have on the readme, and it rounds up to keep things clean :)

Okay coverage seems to be at 99.60% now so hopefully good to go? @daveshanley

TristanSpeakEasy:
I have pull requests ready to go in [libopenapi-validator](https://github.com/pb33f/libopenapi-validator/pull/38) and [vacuum](https://github.com/daveshanley/vacuum/pull/386) once this merges and I can update those dependencies

datamodel/low/v3/create_document.go
daveshanley: This is a problem in vacuum for sure. after A/B testing Stripe:

Before change:
`vacuum took 820 milliseconds to lint 4.28 MB across 1 files`

After change: 
`vacuum took 1.12 seconds to lint 4.28 MB across 1 files`

Same with asana:

Before: 
`vacuum took 217 milliseconds to lint 2.07 MB across 1 files`

After :
`vacuum took 329 milliseconds to lint 2.07 MB across 1 files`

The rolodex is not doing any resolving here, so why do we need to wait?

TristanSpeakEasy: so what we were seeing before this change went in was the yaml.Node structure of the document referred to by the RootNode was being modified while still trying to be parsed to JSON within the extract function, which led to infinite loops within that parsing and then the callstack would blow up.

daveshanley: This looks like a deeper race condition in the resolver or the rolodex. Lets keep this in place as a safety net, until the root cause can be discovered.

daveshanley (approved):
Looks good, I think you need to rebase however, there are conflicts.

Issue #231, opened 2024-01-05
Missing Line and Column Information in model.ChangeContext for Enum Changes.


Amitrei:
Hi there,
We are currently exploring the integration of your remarkable tool into our company\'s workflow. One of the essential functionalities we require involves comparing two OpenAPI documents and presenting the changes within our custom UI, akin to[ the OpenAPI diff](https://github.com/pb33f/openapi-changes) report.

### The issue
The current challenge we face is related to the `model.ChangeContext` object when modifying an enum. Unfortunately, this object does not include information about the new/original line and column, which is crucial for visualizing changes on our UI platform. The issue becomes evident in the following test:

```go
func testEnumChange() { 
	left := `openapi: 3.0
components:
  schemas:
    OK:
      enum: [a,b,c]`

	right := `openapi: 3.0
components:
  schemas:
    OK:
      enum: [a,b,foo]`

	originalDoc, err := libopenapi.NewDocument([]byte(left))
	if err != nil {
		fmt.Println("Failed creating new document from OAS content due to: ", err)
	}
	updatedDoc, err := libopenapi.NewDocument([]byte(right))
	if err != nil {
		fmt.Println("Failed creating new document from OAS content due to: ", err)
	}
	compareResults, _ := libopenapi.CompareDocuments(originalDoc, updatedDoc)

	for _, change := range compareResults.GetAllChanges() {
		fmt.Println("Detected the following change:")

		if change.Original != "" {
			fmt.Printf("property: %v \n original value: %v \n line: %v \n column: %v \n", change.Property, change.Original, *change.Context.OriginalLine, *change.Context.OriginalColumn)
		}

		if change.New != "" {
			fmt.Printf("property: %v \n new value: %v \n line: %v \n column: %v \n", change.Property, change.New, *change.Context.NewLine, *change.Context.NewColumn)
		}

	}

}


```

the results are:
```
Detected the following change:
property: enum 
 new value: foo 
 line: 0 
 column: 0 
Detected the following change:
property: enum 
 original value: c 
 line: 0 
 column: 0 

```

### The cause
This problem stems from the encoding of the `yaml.Node` object in the `toString` method, used [here](https://github.com/pb33f/libopenapi/blob/main/what-changed/model/schema.go#L920-L923).

The `toString` implementation:
```
func toString(v any) string {
	if y, ok := v.(*yaml.Node); ok {
		_ = y.Encode(&v)
	}

	return fmt.Sprint(v)
}
```
[...]

PR #232, opened 2024-01-11
Missing Line and Column Information in model.ChangeContext for Enum Changes fix

Amitrei:
A fix for issue #231 

I have addressed the issue by implementing a simple fix: removing the encoding from the toString function. However, it became apparent that this alone was insufficient. The toString function is crucial when storing the node.Yaml enum object string value as a key in a map, which is used for comparison between the right and left sides. 

To mitigate this issue, I have modified the toString function to return the encoded string without including the column and line information. To prevent the loss of the original object's line and column data, I have utilized a copy of the object. This ensures that the comparison is not adversely affected by the line and column information, while still preserving the integrity of the original object's data.

### DIFF
--- a/what-changed/model/comparison_functions.go
+++ b/what-changed/model/comparison_functions.go
@@ -405,7 +405,9 @@ func ExtractStringValueSliceChanges(lParam, rParam []low.ValueReference[string],
 
 func toString(v any) string {
 	if y, ok := v.(*yaml.Node); ok {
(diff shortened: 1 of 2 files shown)


daveshanley (approved):
Looks good to me! Thank you for your contribution.

daveshanley:
The code coverage issue is a glitch with codecov, all code is covered.

PR #239, opened 2024-01-21
[bundler] Add exploded ref check against root document

### LINKED ISSUE #238: Bundling local files causes root file references to be inlined.
Based on documentation `Local (references inside the root document) references are not bundled, They stay in place.` I am experiencing something different. 

### Actual behavior
```yaml
openapi: 3.1.0
info:
  title: Example
  version: 1.0.0
components:
  responses:
    GetItems:
      content:
        application/json:
          schema:
            type: array
            items:
              $ref: '#/components/schemas/Item'
  schemas:
    Item:
      type: object
      properties:
        name:
          type: string
```
If I bundle this this spec, local references become inlined.
```yaml
openapi: 3.1.0
info:
  title: Example
  version: 1.0.0
components:
  responses:
    GetItems:
      content:
        application/json:
          schema:
            type: array
            items:
              type: object
              properties:
                name:
                  type: string
  schemas:
    Item:
      type: object
      properties:
        name:
          type: string
```
### Expected behavior
I would expect original spec unchanged as it is a standalone ( not exploded ).

### Steps to reproduce
Bundle un-exploded spec file with `BasePath` set.

P.S. Bundling remote references seem to not have this effect.
An example of a diff between remote/local bundles:
```go
package main

import (
	"bytes"
	"errors"
	"fmt"
	"github.com/pb33f/libopenapi"
	"github.com/pb33f/libopenapi/bundler"
	"github.com/pb33f/libopenapi/datamodel"
	"io"
	"net/http"
	"os"
	"os/exec"
	"path/filepath"
)

func main() {
	tmp, _ := os.MkdirTemp("", "petstore")
	defer os.RemoveAll(tmp)

	response, err := http.Get("https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.0/petstore.yaml")
	if err != nil {
		panic(err)
	}

	spec, _ := io.ReadAll(response.Body)
	response.Body.Close()

	remote, err := libopenapi.NewDocumentWithConfiguration(spec, &datamodel.DocumentConfiguration{})
	if err != nil {
		panic(err)
	}

	local, err := libopenapi.NewDocumentWithConfiguration(spec, &datamodel.DocumentConfiguration{
		BasePath: tmp,
	})
	if err != nil {
		panic(err)
	}

	var remoteBundle []byte
	{
		doc, errs := remote.BuildV3Model()
		if err := errors.Join(errs...); err != nil {
			panic(err)
		}

		remoteBundle, err = bundler.BundleDocument(&doc.Model)
		if err != nil {
			panic(err)
		}
	}

	var localBundle []byte
	{
		doc, errs := local.BuildV3Model()
		if err := errors.Join(errs...); err != nil {
			panic(err)
		}

		localBundle, err = bundler.BundleDocument(&doc.Model)
		if err != nil {
			panic(err)
		}
	}

	diff, errs := libopenapi.CompareDocuments(remote, local)
	if err = errors.Join(errs...); err != nil {
		panic(err)
	}

	fmt.Printf("Total changes: %v\n", diff.TotalChanges()) // 0 changes

	if !bytes.EqualFold(remoteBundle, localBundle) { // never equal
		remotePath := filepath.Join(tmp, "remotebundle")
		os.WriteFile(remotePath, remoteBundle, 0644)

		localPath := filepath.Join(tmp, "localbundle")
		os.WriteFile(localPath, localBundle, 0644)

		out, err := exec.Command("diff", "-u", remotePath, localPath).CombinedOutput()
		if err != nil {
			switch err.(type) {
			case *exec.ExitError:
			default:
				panic(err)
			}
		}

		fmt.Println(string(out))
	}
}
```

Domarx:
Exploded ref definition was matching against empty string, however full definition consists of [full path and a fragment](https://github.com/pb33f/libopenapi/blob/main/index/extract_refs.go#L219)

Fixes #238

### DIFF
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -58,7 +58,7 @@ func bundle(model *v3.Document, inline bool) ([]byte, error) {
 			//if we're in the root document, don't bundle anything.
 			refExp := strings.Split(sequenced.FullDefinition, "#/")
 			if len(refExp) == 2 {
-				if refExp[0] == sequenced.Index.GetSpecAbsolutePath() {
+				if refExp[0] == sequenced.Index.GetSpecAbsolutePath() || refExp[0] == "" {
 					if root && !inline {
 						idx.GetLogger().Debug("[bundler] skipping local root reference",
 							"ref", sequenced.Definition)
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -58,7 +58,7 @@ func bundle(model *v3.Document, inline bool) ([]byte, error) {
 			//if we're in the root document, don't bundle anything.
 			refExp := strings.Split(sequenced.FullDefinition, "#/")
 			if len(refExp) == 2 {
-				if refExp[0] == "" {
+				if refExp[0] == sequenced.Index.GetSpecAbsolutePath() {
 					if root && !inline {
 						idx.GetLogger().Debug("[bundler] skipping local root reference",
 							"ref", sequenced.Definition)
--- a/bundler/bundler_test.go
+++ b/bundler/bundler_test.go
@@ -5,6 +5,7 @@ package bundler
 
 import (
 	"bytes"
+	"errors"
 	"github.com/pb33f/libopenapi"
 	"github.com/pb33f/libopenapi/datamodel"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


bundler/bundler.go
```diff
@@ -58,7 +58,7 @@ func bundle(model *v3.Document, inline bool) ([]byte, error) {
 			//if we're in the root document, don't bundle anything.
 			refExp := strings.Split(sequenced.FullDefinition, "#/")
 			if len(refExp) == 2 {
-				if refExp[0] == "" {
+				if refExp[0] == sequenced.Index.GetSpecAbsolutePath() {
```
daveshanley, line 61: Sometimes (depending on the configuration) there may or may not be a full path, it may be empty (which is why the original logic stands). 

```suggestion
				if refExp[0] == sequenced.Index.GetSpecAbsolutePath() || refExp[0] == "" {
```

daveshanley (changes requested):
If you add the suggested code, I can merge this in.

daveshanley (approved):
Looks good to me! Thank you.

PR #241, opened 2024-01-25
Fix circular polymorfic ref

eli-l:
- Test case taken from [docs](https://pb33f.io/libopenapi/circular-references/#circular-reference-results)
- Applying fix for _TestResolver_CheckForCircularReferences_IgnorePoly_Any_ to pass
- Adjusting expectations for _TestRolodex_IndexCircularLookup_LookupHttpNoBaseURL_  because of **cf.IgnorePolymorphicCircularReferences = true**


daveshanley (approved):
Wow, fixing a bug at this level deserves a round of applause! This is some complex code that no-one goes near normally. 

Well done!

PR #244, opened 2024-01-29
Addressed invalid polymorphic circular reference issue.

daveshanley:
If a polymorphic type was used incorrectly (as a map and not an array) then the logic was getting a little upset. This fixes that issue and adds some tests to validate correct handling.


TristanSpeakEasy (approved):
LGTM

PR #260, opened 2024-02-25
fix bundler panic when BuildV3Model fail

emilien-puget:
in some cases, `BuildV3Model` will return a nil first parameter and an error which result in a panic in the next call, this PR aims to fix this behavior.

### DIFF
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -14,6 +14,9 @@ import (
 	"github.com/pb33f/libopenapi/index"
 )
 
+// ErrInvalidModel is returned when the model is not usable.
+var ErrInvalidModel = errors.New("invalid model")
+
 // BundleBytes will take a byte slice of an OpenAPI specification and return a bundled version of it.
 // This is useful for when you want to take a specification with external references, and you want to bundle it
 // into a single document.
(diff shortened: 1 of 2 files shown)


emilien-puget:
there is no relevant way to increase the code coverage for this change

daveshanley:
> there is no relevant way to increase the code coverage for this change

What about changing the design so that a `failOnError` bool can be passed in to decide if it should return on any build errors, or continue going. 

The reason why the coverage is failing is because it should not be a 'any failure results in total failure' scenario. 

There could be many reasons why specific schemas failed to build or things went wrong deeper in the build tree, that is why there is an error returned. Just because one thing failed, it does not mean the rest of the model is toast. If there is a model, then you are good to go. 

Here is where we would allow the consumer to decide if they want to fail at this point (return the error) or continue on.

To continue, would allow the circular reference check to pass (which is why errors may have returned, but the model knows how to handle them when building).

emilien-puget:
we could return when the model is nil instead of when any error is returned, allowing the consumer to generate a panic is not great

emilien-puget:
@daveshanley i changed the error handling to be more in sync with the previous behavior, while still being panic free

daveshanley (approved):
LGTM!

Issue #269, opened 2024-04-04
Remote reference cannot be resolved


yozel:
Hi, I found out a bug with the remote references.

Steps to reproduce:
```go
package main

import (
	"errors"
	"fmt"

	"github.com/pb33f/libopenapi"
	"github.com/pb33f/libopenapi/datamodel"
)

var spec = []byte(`openapi: "3.0.0"
info:
  title: test
  version: "3"
paths: { }
components:
  schemas:
    Container:
      properties:
        pet:
          $ref: https://petstore3.swagger.io/api/v3/openapi.json#/components/schemas/Pet
`)

func main() {
	doc, err := libopenapi.NewDocumentWithConfiguration(spec, &datamodel.DocumentConfiguration{
		AllowRemoteReferences: true,
	})
	if err != nil {
		panic(err)
	}
	_, errs := doc.BuildV3Model()
	if len(errs) > 0 {
		fmt.Println(errors.Join(errs...))
	}
}
```

```
go run ./issues/...
cannot resolve reference `https://petstore3.swagger.io/api/v3/openapi.json#/components/schemas/Pet`, it's missing: $https:.['petstore3.swagger.io'].api.v3['openapi.json$'].components.schemas.Pet [11:11]
```

According to `git bisect` the culprit is https://github.com/pb33f/libopenapi/commit/7119b513366b187d605cb9d6c8e4886e3df57a1f, which I could confirm that reverting this commit actually solves the issue.

daveshanley:
The commit was to prevent non http lookups from hitting the remote file system code, but it seems to have had a strange effect, thanks for finding this - will look into it.

daveshanley:
This regression has been fixed in `v0.16.0`

A test was added to validate: https://github.com/pb33f/libopenapi/blob/main/document_test.go#L1348

PR #274, opened 2024-04-18
v0.15.16

daveshanley:
Addresses issues: 

- #269


daveshanley:
For reference, the missing line is actually covered by the test as it executes, for some reason the coverage tool is not picking it up, but the code does run, it's what caused a regression in the first place.

lesson: only place to much trust in tooling.

PR #276, opened 2024-04-22
fix: fixed MediaType example being populated from wrong source

TristanSpeakEasy:
@daveshanley this fixes an issue where the `MediaType.Example` field was being populated when it shouldn't.

It was basically searching down into the schema and other places as well to find an example where it should have only been searching the top level node


daveshanley:
Cool! for some reason codecov has stopped working, trying to shake it loose.

daveshanley:
Looks good! thank you for fixing.

PR #287, opened 2024-05-07
Pin digital ocean commit in tests

maboehm:
I wanted to open a PR for something else, but the tests were failing.

The digital ocean API repo does not change often, but when it does it likely breaks the tests here.
I'd suggest to pin a certain commit.

If you want, I can also de-duplicate the checkout code between the tests, but it looks like the tests were explicit on purpose


daveshanley:
Thanks for pinning them, it was getting annoying updating the tests all the time.

PR #288, opened 2024-05-07
Fix RenderAndReload error handling

maboehm:
~~(unrelated tests are currently failing, fixed in #287 )~~ ✅ 

The error handling for `newDoc.BuildV3Model()` was wrong and the error was not properly propagated. The provided test only runs successful with the change.


daveshanley:
tests should be working again after your other PR was merged.

maboehm:
@daveshanley I am not sure how I can write a test that would trigger that error. In my opinion the code coverage is a bit too picky here

daveshanley:
Fair enough, I don't know how to write a test to trigger that error either.

Issue #289, opened 2024-05-08
Bug: securitySchemes oauth2 empty scopes is stripped


khart-twilio:
input:
```
components:
  securitySchemes:
    someName:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: https://oauth.somewhere.com/v1/token
          scopes: {}
```
invalid output (`scopes` key is removed but is required as per https://swagger.io/docs/specification/authentication/oauth2/):
```
components:
  securitySchemes:
    someName:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: https://oauth.somewhere.com/v1/token
```

can this be fixed, and in the meantime what is the best way to work around the issue?

Thanks for the library, I love the website style :)

daveshanley:
Fixed in `v0.16.6` Thank you for your contribution!

khart-twilio:
Thanks, I was digging into the tests but it takes a while to get familiar.  I appreciate you taking it up!

PR #290, opened 2024-05-09
render empty oauth scopes

khart-twilio:
see https://github.com/pb33f/libopenapi/issues/289

This is my naive attempt to fix the issue.  It works but it may have negative side-effects I'm not aware of.

### DIFF
--- a/orderedmap/builder.go
+++ b/orderedmap/builder.go
@@ -36,6 +36,9 @@ type findValueUntyped interface {
 // ToYamlNode converts the ordered map to a yaml node ready for marshalling.
 func (o *Map[K, V]) ToYamlNode(n NodeBuilder, l any) *yaml.Node {
 	p := utils.CreateEmptyMapNode()
+	if o != nil {
+		p.Content = make([]*yaml.Node, 0)
+	}
 
 	var vn *yaml.Node
 
--- a/datamodel/high/node_builder.go
+++ b/datamodel/high/node_builder.go
@@ -429,7 +429,7 @@ func (n *NodeBuilder) AddYAMLNode(parent *yaml.Node, entry *nodes.NodeEntry) *ya
 			}
 
 			p := m.ToYamlNode(n, l)
-			if len(p.Content) > 0 {
+			if p.Content != nil {
 				valueNode = p
 			}
 		} else if r, ok := value.(Renderable); ok {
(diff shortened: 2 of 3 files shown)


orderedmap/builder.go
```diff
@@ -36,6 +36,9 @@ type findValueUntyped interface {
 // ToYamlNode converts the ordered map to a yaml node ready for marshalling.
 func (o *Map[K, V]) ToYamlNode(n NodeBuilder, l any) *yaml.Node {
 	p := utils.CreateEmptyMapNode()
+	if o != nil {
```
daveshanley, line 39: Can you add a small test for this, as this new block will reduced coverage, if you pull the project again and update the workflow config (upgraded the codecov action) You should see the coverage on this PR.

khart-twilio:
@daveshanley One edge case I noticed when testing:

I had: 
```
...
      requestBody:
        content:
          application/json:
            schema:
              type: array
              items:
                type: object
                properties:
                  value:
                    type: null # null isn't allowed here I don't think according to the spec
```
And with these changes in libopenapi it is now being processed and spit out as

```
...
                properties:
                  value:
                    type: "ObjectNode"
```

It doesn't matter for this case since using `null` was invalid and I changed it to something else, but I'm wondering if it might matter in other places.  I'm not familiar with how things like `null`, `true`, `false` come through as yaml.Node object when deserialized and if I'd need to handle these cases in the changes.

orderedmap/builder.go
khart-twilio: I merged the latest change from main branch and ran `go test -v -coverprofile cover.out ./...` locally and see 100% coverage.  In the status on this PR I see `1 workflow awaiting approval`, which I assume means you need to approve something before the coverage check can run here, but I'm unfamiliar.  I'd appreciate if you clarify for me.

khart-twilio: whoops, nevermind I was reading the test output incorrectly.  I'll try to figure it out.

daveshanley:
> @daveshanley One edge case I noticed when testing:
> 
> I had:
> 
> ```
> ...
>       requestBody:
>         content:
>           application/json:
>             schema:
>               type: array
>               items:
>                 type: object
>                 properties:
>                   value:
>                     type: null # null isn't allowed here I don't think according to the spec
> ```
> 
> And with these changes in libopenapi it is now being processed and spit out as
> 
> ```
> ...
>                 properties:
>                   value:
>                     type: "ObjectNode"
> ```
> 
> It doesn't matter for this case since using `null` was invalid and I changed it to something else, but I'm wondering if it might matter in other places. I'm not familiar with how things like `null`, `true`, `false` come through as yaml.Node object when deserialized and if I'd need to handle these cases in the changes.

I have no idea where `ObjectNode` is coming from. It's not part of `libopenapi` 

Because I have no idea, let's just proceed, if someone else finds this issue then we can investigate further.

Issue #285, opened 2024-05-02
 Panic During Schema Rendering


matteogastaldello:
Hi! 
I'm encountering a runtime panic due to an invalid memory address or nil pointer dereference when attempting to render a schema using the libopenapi library. This issue arises during the execution of the schema.Render() method, which results in a segmentation violation error.

These are the lines that throws the panic
```
	schema, err = bodySchema.BuildSchema()
	if err != nil {
		return fmt.Errorf("building schema for %s: %w", verb.Path, err), errors
	}
	schema.Properties.Set("authenticationRefs", base.CreateSchemaProxy(&base.Schema{
		Type:        []string{"object"},
		Description: "AuthenticationRefs represent the reference to a CR containing the authentication information. One authentication method must be set."}))
	schema.Required = append(schema.Required, []string{"authenticationRefs"}...)
	rend, err := schema.Render()
	if err != nil {
		return fmt.Errorf("rendering schema for %s: %w", verb.Path, err), errors
	}
```

This error occurs if schema.Required is not previously set in the oas specification used to build the model. Setting a "fake" required field in the oas specification file fix the behaviour.

Thanks for your help

daveshanley:
Hi, 

What will really help me diagnose this problem is a blob of code I can copy and paste into my test-rig and debug the issue, the problem with snippets like this is that I have no idea how your source is structured or your base specification etc. 

I really don't have the luxury of time to re-create this issue by reverse engineering the code you've provided. I really need something that will re-create the issue, that runs as a test or that will run as a copy paste job. 

Thanks.

matteogastaldello:
Hi @daveshanley!

Here is the gist with all the information to reproduce the error: https://gist.github.com/matteogastaldello/823cdaf0127c2f2eaa92ddb574539825

Thanks a lot!

daveshanley:
Resolved in `v0.16.9`

PR #300, opened 2024-06-06
v0.16.9

daveshanley:
Adds `GetKeyNode` and `GetRootNode` methods to low level models. Useful for interfaces in downstream applications.
A general tuneup in various places after hammering the library as part of a larger stack of applications. 

Addresses issues:

- #285


daveshanley:
I cannot bump the coverage in the node builder. Overall overage is lifted by 0.06%

Issue #301, opened 2024-06-13
Panic when rendering a nullable example


ThomasRooney:
Given a specification (e.g. stored under `test_specs/nullable-examples.openapi.yaml`:

```
openapi: 3.1.0
components:
  schemas:
    Thing:
      type: object
      description: A nullable example.
      properties:
        target:
          nullable: true
          type: string
          enum:
            - staging
            - production
          example:
```

Attempting to render this specification causes a panic. 

```
func TestDocument_Nullable_Example(t *testing.T) {
	data, _ := os.ReadFile("../../../test_specs/nullable-examples.openapi.yaml")
	info, _ := datamodel.ExtractSpecInfo(data)

	lowDoc, _ = lowv3.CreateDocumentFromConfig(info, datamodel.NewDocumentConfiguration())

	highDoc := NewDocument(lowDoc)
	rendered := highDoc.RenderWithIndention(2)

	if runtime.GOOS != "windows" {
		assert.Equal(t, string(data), strings.TrimSpace(string(rendered)))
	}

	rendered = highDoc.RenderWithIndention(4)
	if runtime.GOOS != "windows" {
		assert.NotEqual(t, string(data), strings.TrimSpace(string(rendered)))
	}
}
```

```
=== RUN   TestDocument_Nullable_Example
--- FAIL: TestDocument_Nullable_Example (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10439b798]

goroutine 35 [running]:
testing.tRunner.func1.2({0x1048a6580, 0x104d2a520})
	/opt/homebrew/opt/go/libexec/src/testing/testing.go:1631 +0x3c0
testing.tRunner.func1()
	/opt/homebrew/opt/go/libexec/src/testing/testing.go:1634 +0x57c
panic({0x1048a6580?, 0x104d2a520?})
	/opt/homebrew/opt/go/libexec/src/runtime/panic.go:770 +0xf0
gopkg.in/yaml%2ev3.handleErr(0x1400028fc38)
	/Users/thomasrooney/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/yaml.go:294 +0x120
[...]

PR #302, opened 2024-06-13
fix: render empty scalar nodes

ThomasRooney:
Fixes https://github.com/pb33f/libopenapi/issues/301.

TL;DR go-yaml currently panics when trying to render the test case. So we've got to short-circuit the "empty node" to not use `Encode` to copy across the empty scalar node, but just set the value explicitly


daveshanley (approved):
LGTM! Will add to the next release.

PR #303, opened 2024-06-15
removed golang x package in favour of the std lib

emilien-puget:
removed golang.org/ x package in favour of the std lib


emilien-puget:
after changing x/sync to sync, the import order was weird si ran gci to re order them, turns out import order was not inforced

daveshanley:
Great! Thank you. Will merge this in with the next release.

PR #310, opened 2024-07-22
Update Speakeasy URL

ndimares:
Speakeasy is now found at speakeasy.com!


daveshanley (approved):
Nice! I am sure that cost a few bucks ;)

PR #312, opened 2024-07-30
fix: fixed extraction of top level named examples to not extract schema level examples

TristanSpeakEasy:
This fixes an issues where the schema level examples were wrongly being extracted and didn't properly match the shape of the top level named examples.

This PR ensures only the right examples are extracted


TristanSpeakEasy:
@daveshanley would appreciated a review

daveshanley (approved):
Cool, LGTM!

PR #316, opened 2024-08-05
fix: improved handling of complex yaml in YAMLNodeToJSON

TristanSpeakEasy:
Makes it possible to render yaml that uses anchors and complex map keys

### DIFF
--- a/json/json.go
+++ b/json/json.go
@@ -3,14 +3,13 @@ package json
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/pb33f/libopenapi/orderedmap"
 	"gopkg.in/yaml.v3"
 )
 
 // YAMLNodeToJSON converts yaml/json stored in a yaml.Node to json ordered matching the original yaml/json
-//
-// NOTE: The limitation is this won't work with YAML that is not compatible with JSON, ie yaml with anchors or complex map keys
 func YAMLNodeToJSON(node *yaml.Node, indentation string) ([]byte, error) {
 	v, err := handleYAMLNode(node)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


daveshanley:
Bump those two lines of coverage, and it's good to go.

json/json.go
```diff
@@ -31,28 +30,38 @@ func handleYAMLNode(node *yaml.Node) (any, error) {
...
-	m := orderedmap.New[string, yaml.Node]()
+	v := orderedmap.New[string, any]()
+	for i, n := range node.Content {
+		if i%2 == 0 {
+			continue
+		}
+		keyNode := node.Content[i-1]
+		kv, err := handleYAMLNode(keyNode)
+		if err != nil {
+			return nil, err
```
daveshanley, line 48: Coverage?

json/json.go
```diff
@@ -31,28 +30,38 @@ func handleYAMLNode(node *yaml.Node) (any, error) {
...
+		}
 
-	if err := node.Decode(m); err != nil {
-		return nil, err
-	}
+		if reflect.TypeOf(kv).Kind() != reflect.String {
+			keyData, err := json.Marshal(kv)
+			if err != nil {
+				return nil, err
```
daveshanley, line 54: Coverage?

TristanSpeakEasy:
@daveshanley essential asking for a pass on dealing with those bits of coverage, though if you really do want them covered I guess a way forward is to unit test the internal functions directly (currently I am only testing the public interface with the `package json_test` restriction in the test file) and manually trying to construct an invalid yaml.Node

daveshanley:
I spent about 20 minutes and I could get one line covered, but the other is impossible to test as it's a deep catch. 

Pass granted.

PR #319, opened 2024-08-14
chore: update to use iterators on orderedmaps

TristanSpeakEasy:
This PR adds support for the iterators being added to the ordered map library, this cleans up the DevEx when working with ordered maps in the high level models but has also cleaned up a bunch of the implementation code as well

I have linked to my fork of the iterators change for now, but I can hold of this PR (just let me know) until its merged into the library

https://github.com/wk8/go-ordered-map/pull/41

Just some notes about things I noticed while doing this refactor:

- There is a lot of duplicated code that could be reduced with some helper methods, I added some but stopped short of going to far in one PR
- The way maps are hashed is very inconsistent and in a lot of cases changing key names (as long as they don't change the order of the sorted map) wouldn't be caught in a hash change, worth probably doing a pass on this in a separate PR

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
-          go-version: 1.21
+          go-version: 1.23
         id: go
 
       - name: Checkout code
@@ -35,7 +35,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
-          go-version: 1.21
+          go-version: 1.23
         id: go
 
       - name: Checkout code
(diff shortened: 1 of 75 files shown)


TristanSpeakEasy:
a draft until the underlying dependency is merged otherwise we either need to use speakeasy's fork without a replace statement or everyone needing to use this version has to add their own replace statement

daveshanley:
I will need to look at this in depth, thank you for this work!

TristanSpeakEasy:
> I will need to look at this in depth, thank you for this work!

@daveshanley we have integrated this branch into the speakeasy products and its working as expected and much nicer devex

TristanSpeakEasy:
no longer using a replace directive so good to go

petkostas:
I started testing the specific PR, and it looks like it is working as expected! Thanks a lot, @TristanSpeakEasy

daveshanley (approved):
This looks great. A very nice upgrade.

PR #322, opened 2024-08-16
Use real root filename as theoretical root yaml file

lobocv:
Instead of using "root.yaml" as a theoretical root, use the real filename of the root spec (often openapi.yaml). This fixes issues with resolving references in rules such as unused-components.

### DIFF
--- a/index/index_model.go
+++ b/index/index_model.go
@@ -87,6 +87,9 @@ type SpecIndexConfig struct {
 	// If resolving locally, the BasePath will be the root from which relative references will be resolved from
 	BasePath string // set the Base Path for resolving relative references if the spec is exploded.
 
+	// SpecFilePath is the name of the root specification file (usually named "openapi.yaml").
+	SpecFilePath string
+
 	// In an earlier version of libopenapi (pre 0.6.0) the index would automatically resolve all references
 	// They could have been local, or they could have been remote. This was a problem because it meant
 	// There was a potential for a remote exploit if a remote reference was malicious. There aren't any known
--- a/index/index_model.go
+++ b/index/index_model.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"path/filepath"
 	"sync"
 
 	"github.com/pb33f/libopenapi/datamodel"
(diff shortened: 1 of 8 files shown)


index/rolodex.go
```diff
@@ -425,6 +426,26 @@ func (r *Rolodex) BuildIndexes() {
 	r.manualBuilt = true
 }
 
+// GetAllReferences  returns all references found in the root and all other indices
+func (r *Rolodex) GetAllReferences() map[string]*Reference {
+	allRefs := make(map[string]*Reference)
+	for _, idx := range append(r.GetIndexes(), r.GetRootIndex()) {
```
lobocv, line 432: @daveshanley I made these methods because using GetIndexes() alone was not enough. It seems that GetIndexes() does not include the root index. Is that on purpose? Rather than returning the root index in GetIndexes(), I added these methods, just to be safe.

index/spec_index.go
```diff
@@ -152,6 +156,13 @@ func (index *SpecIndex) GetRolodex() *Rolodex {
 	return index.rolodex
 }
 
+func (index *SpecIndex) GetSpecFileName() string {
```
daveshanley, line 159: Can you add a comment here for docs

index/index_model.go
```diff
@@ -151,6 +155,17 @@ type SpecIndexConfig struct {
 	uri []string
 }
 
+// SetTheoreticalRoot sets the spec file paths to a theoretical file path
```
daveshanley, line 158: Can you just add a little more here to explain what this means for folks who don't know why we added this?

index/rolodex_test.go
```diff
@@ -1516,18 +1516,27 @@ func TestRolodex_SimpleTest_OneDoc(t *testing.T) {
 	}
 
 	cf := CreateOpenAPIIndexConfig()
+	cf.SpecFilePath = filepath.Join(baseDir, "doc1.yaml")
```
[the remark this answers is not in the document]
daveshanley: It's a complex test to validate deeply nested files and used as a test for debugging lookup issues. I required it when building out the rolodex last year.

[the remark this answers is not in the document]
lobocv: But is is supposed to have two root spec files (doc1 and doc2)? Seems strange. Are my changes okay? I picked one of them to be the root

daveshanley (approved):
LGTM

PR #327, opened 2024-09-03
Fix: resolve remote relative refs correctly

felixjung:
Thanks for working on all these tools for OpenAPI. 👏 

I've run into issues with vacuum/libopenapi supporting resolving of relative remote refs. I'm trying to fix this, but I have to admit I get completely lost in the code of this project.

I've added a test case to debug the issue. One thing I notice and that I think it should never happen, is that `libopenapi` tries to load a file that's not referenced anywhere. The log output says the following:

```
{"time":"2024-09-03T22:46:30.36193+02:00","level":"ERROR","msg":"unable to fetch remote document","file":"/felixjung/libopenapi/authed-remote/test_specs/schemas/components.openapi.yaml","status":404,"resp":"404: Not Found"}
{"time":"2024-09-03T22:46:30.362202+02:00","level":"ERROR","msg":"unable to locate reference anywhere in the rolodex","reference":"https://raw.githubusercontent.com/felixjung/libopenapi/authed-remote/test_specs/schemas/components.openapi.yaml#/components/responses/ListAccounts"}
{"time":"2024-09-03T22:46:30.362272+02:00","level":"ERROR","msg":"error building path item: map build failed: reference cannot be found: reference 'https://raw.githubusercontent.com/felixjung/libopenapi/authed-remote/test_specs/schemas/components.openapi.yaml#/components/responses/ListAccounts' at line 31, column 11 was not found"}
```

Instead of

```
https://raw.githubusercontent.com/felixjung/libopenapi/authed-remote/test_specs/schemas/components.openapi.yaml#/components/responses/ListAccounts
```

it should only ever try to resolve

```
https://raw.githubusercontent.com/felixjung/libopenapi/authed-remote/test_specs/minimal_remote_refs/schemas/components.openapi.yaml#/components/responses/ListAccounts
```

When I call a Render method on the document's model, components are missing and refs have not been resolved (maybe I'm using it incorrectly).

To run the test do

```
go test -count 1 -v -run TestDocument_MinimalRemoteRefs .
```

I'd be very appreciative for any hints as to what might be going wrong.

### DIFF
--- a/datamodel/low/extraction_functions.go
+++ b/datamodel/low/extraction_functions.go
@@ -87,7 +87,6 @@ func LocateRefNodeWithContext(ctx context.Context, root *yaml.Node, idx *index.S
 		for _, collection := range collections {
 			found = collection()
 			if found != nil && found[rv] != nil {
-
 				// if this is a ref node, we need to keep diving
 				// until we hit something that isn't a ref.
 				if jh, _, _ := utils.IsNodeRefValue(found[rv].Node); jh {
@@ -147,7 +146,7 @@ func LocateRefNodeWithContext(ctx context.Context, root *yaml.Node, idx *index.S
 								u := *idx.GetConfig().BaseURL
 								p := ""
 								if u.Path != "" {
-									p = filepath.Dir(u.Path)
+									p = u.Path
 								}
 								u.Path = filepath.Join(p, explodedRefValue[0])
 								rv = fmt.Sprintf("%s#%s", u.String(), explodedRefValue[1])
--- a/datamodel/low/extraction_functions.go
+++ b/datamodel/low/extraction_functions.go
@@ -105,49 +105,75 @@ func LocateRefNodeWithContext(ctx context.Context, root *yaml.Node, idx *index.S
 			}
 		}
 
-		// perform a search for the reference in the index
-		// extract the correct root
+		// Obtain the absolute filepath/URL of the spec in which we are trying to
+		// resolve the reference value [rv] from. It's either available from the
+		// index or passed down through context.
 		specPath := idx.GetSpecAbsolutePath()
 		if ctx.Value(index.CurrentPathKey) != nil {
 			specPath = ctx.Value(index.CurrentPathKey).(string)
 		}
 
+		// explodedRefValue contains both the path to the file containing the
+		// reference value at index 0 and the path within that file to a specific
+		// sub-schema, should it exist, at index 1.
 		explodedRefValue := strings.Split(rv, "#")
 		if len(explodedRefValue) == 2 {
+			// The ref points to a component within either this file or another file.
 			if !strings.HasPrefix(explodedRefValue[0], "http") {
+				// The ref is not an absolute URL.
 				if !filepath.IsAbs(explodedRefValue[0]) {
+					// The ref is not an absolute local file path.
 					if strings.HasPrefix(specPath, "http") {
+						// The schema containing the ref is itself a remote file.
 						u, _ := url.Parse(specPath)
+						// p is the directory the referenced file is expected to be in.
 						p := ""
 						if u.Path != "" && explodedRefValue[0] != "" {
+							// We are using the path of the resolved URL from the rolodex to
+							// obtain the "folder" or base of the file URL.
 							p = filepath.Dir(u.Path)
 						}
 						if p != "" && explodedRefValue[0] != "" {
+							// We are resolving the relative URL against the absolute URL of
+							// the spec containing the reference.
 							u.Path = utils.ReplaceWindowsDriveWithLinuxPath(filepath.Join(p, explodedRefValue[0]))
 						}
 						u.Fragment = ""
+						// Turn the reference value [rv] into the absolute filepath/URL we
+						// resolved.
 						rv = fmt.Sprintf("%s#%s", u.String(), explodedRefValue[1])
-
 					} else {
+						// The schema containing the ref is a local file or doesn't have an
+						// absolute URL.
 						if specPath != "" {
+							// We have _some_ path for the schema containing the reference.
 							var abs string
 							if explodedRefValue[0] == "" {
+								// Reference is made within the schema file, so we are using the
+								// same absolute local filepath.
 								abs = specPath
 							} else {
 								// break off any fragments from the spec path
 								sp := strings.Split(specPath, "#")
+								// Create a clean (absolute?) path to the file containing the
+								// referenced value.
 								abs, _ = filepath.Abs(filepath.Join(filepath.Dir(sp[0]), explodedRefValue[0]))
 							}
 							rv = fmt.Sprintf("%s#%s", abs, explodedRefValue[1])
 						} else {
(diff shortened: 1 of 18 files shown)


daveshanley:
This one is simple. it's a bug, but can be worked around, and the bugfix is simple. Add a '/' to the end of your URL, i.e : 

`baseURL, err := url.Parse("https://raw.githubusercontent.com/felixjung/libopenapi/authed-remote/test_specs/minimal_remote_refs/")`

https://github.com/pb33f/libopenapi/blob/main/datamodel/low/extraction_functions.go#L150 <-- this line is ripping off minimal_remote_refs because it's trying to create a filepath. the last segment of the path looks like a file (`minimal_remote_refs`)

To patch the code:

```go
if u.Path != "" {
    if u.Path[len(u.Path)-1:] == "/" {
        p = filepath.Dir(u.Path)
    } else {
        p = u.Path
    }
}
```

daveshanley:
> Thanks for working on all these tools for OpenAPI. 👏
> 
> I've run into issues with vacuum/libopenapi supporting resolving of relative remote refs. I'm trying to fix this, but I have to admit I get completely lost in the code of this project.

It's a little over-engineered to cater to a ton of feature work I need in the future, but not today, so it's heavy duty - but because it will need to be down the line. :)

felixjung:
I was able to fix the bug, I guess. What I'm struggling with now is how to write a test that actually fails before committing my fix. Why would building the document or bundling the spec not fail, if a reference cannot be resolved? In this case the request to GitHub to download the file will fail without the fix and that should fail the bundling or document building process. 🤔

felixjung:
I've added a bundler test case, which revealed another issue in that the local reference in the file referenced from the root is not included in the bundle result.

felixjung:
Interestingly, the following test passes just fine:
```go
func TestBundleDocument_MinimalRemoteRefs(t *testing.T) {
	newRemoteHandlerFunc := func() utils.RemoteURLHandler {
		c := &http.Client{
			Timeout: time.Second * 120,
		}

		return func(url string) (*http.Response, error) {
			resp, err := c.Get(url)
			if err != nil {
				return nil, fmt.Errorf("fetch remote ref: %v", err)
			}

			return resp, nil
		}
	}

	specBytes, err := os.ReadFile("../test_specs/minimal_remote_refs/openapi.yaml")
	require.NoError(t, err)

	require.NoError(t, err)

	config := &datamodel.DocumentConfiguration{
		AllowFileReferences:   true,
		AllowRemoteReferences: false,
		BundleInlineRefs:      false,
		BasePath:              "../test_specs/minimal_remote_refs",
		BaseURL:               nil,
		RemoteURLHandler:      newRemoteHandlerFunc(),
	}
	require.NoError(t, err)

	bytes, e := BundleBytes(specBytes, config)
	assert.NoError(t, e)
	assert.Contains(t, string(bytes), "Name of the account", "should contain all reference targets")
}
```

What I've found, trying to debug this, is that there is only one index for the root file. There is no index for the referenced file, despite it containing references to components.

index/rolodex_remote_loader.go
```diff
@@ -462,5 +459,10 @@ func (i *RemoteFS) Open(remoteURL string) (fs.File, error) {
 			i.rolodex.AddExternalIndex(idx, remoteParsedURL.String())
 		}
 	}
+
+	// remove from processing
+	processingWaiter.file = remoteFile
+	processingWaiter.done = true
+	i.ProcessingFiles.Delete(remoteParsedURL.Path)
```
felixjung, line 466: @daveshanley: is this maybe what caused the race condition with not sequentially resolving refs, you mentioned? I was thinking that before this change, a remote file would be marked as done and deleted from `ProcessingFiles` before the index was built. This seems racy to me. 🤔

index/rolodex.go
```diff
@@ -159,14 +160,14 @@ func (r *Rolodex) SetRootNode(node *yaml.Node) {
 
 func (r *Rolodex) AddExternalIndex(idx *SpecIndex, location string) {
 	r.indexLock.Lock()
+	r.indexes = append(r.indexes, idx)
```
felixjung, line 163: This was the bug it took me so long to find. The remote loader uses `AddExternalIndex` to add the index of any remote file to the rolodex. Previously, this wouldn't get the index added to `r.indexes`. When bundling, `rolodex.GetIndexes` would get called but not return the indices of remote files. That's why any refs made from remote files would not get added. 🤦

index/rolodex_remote_loader.go
daveshanley: Very possibly yes. I am sure that waiter code could be more effectively implemented. It took me a while to stop things from locking permanently in there with strange conditions.

index/rolodex.go
daveshanley: Ah! man, I know how mad this code is and how complex it can get when chasing down a bug in there. Well done!

felixjung:
Cleaned up the comment history and this should now be ready to go.

datamodel/low/extraction_functions.go
```diff
@@ -106,49 +105,75 @@ func LocateRefNodeWithContext(ctx context.Context, root *yaml.Node, idx *index.S
...
+							// relative references from. This likely happens when the schema
+							// is the root schema, i.e. the file given to libopenapi as entry.
+							//
 
+							// check for a config BaseURL and use that if it exists.
+							if idx.GetConfig().BaseURL != nil {
 								u := *idx.GetConfig().BaseURL
 								p := ""
 								if u.Path != "" {
-									p = filepath.Dir(u.Path)
```
felixjung, line 150: In a previous version of this PR I conditionally removed the last path segment if it was a file. I reverted this because I believe that we shouldn't make too many assumptions about the URL someone provides as `BaseURL`. Whether it contains a period or not does not mean anything with regards to a URLs content, even `/some/path/openapi.yaml` can in theory have a sub-path yielding`/some/path/openapi.yaml/components.yaml`. It all depends on what the server returns for a remote ref. 🤷 

So, instead this PR now documents what the `BaseURL` should look like and strips any trailing slashes from its path.

datamodel/low/extraction_functions.go
```diff
@@ -106,49 +105,75 @@ func LocateRefNodeWithContext(ctx context.Context, root *yaml.Node, idx *index.S
 			}
 		}
 
-		// perform a search for the reference in the index
-		// extract the correct root
+		// Obtain the absolute filepath/URL of the spec in which we are trying to
```
felixjung, line 108: You OK with leaving these doc comments? I put these here when initially stepping through the code. I hope they'd help me or someone else in the future.

felixjung:
Of course I didnn't see the failing test locally. Damn you, Windows.

daveshanley (approved):
I love all the new comments in `extraction_functions` That block of code is so hard to understand because it handles so many variations and edge cases. (there are others like it in the resolver and the index btw). 

Thank you for digging into this code, it's over-engineered for a reason and getting down and dirty with it, can leave you fried. 

Thank you for your contribution!

PR #336, opened 2024-10-04
fix: a parameter in the rolodex should be evaluated using the index that the node is in

ThomasRooney:
This fixes vacuum rules that lookup operations in the index where:

1. The original spec (root.yaml) has an external reference to a file in a different directory `./paths/mypath.yaml`
2. In that location, there is a relative reference to a path-item in a nearby file, e.g. `./components.yaml#/path-item`

What was previously happening in [2] is that the `./components.yaml` was being resolved in `./components.yaml` (the parent directory) instead of `./paths/components.yaml` (the right directory). This resulted in a failure to evaluate path item references and unexpected linting failures. 

This changes adjusts the lookup to search through the index's nodeMap to find the location of the YAML `$ref` node value that's been evaluated.  Once we find it, we lookup the associated index of that file and do a seek in that specific index.

We also fix a race condition with resolving the specification asyncronously -- if we are going to an external ref we must do it sequentially as otherwise an index might be added multiple times during an async `FindComponent` call. This latent bug was exposed as we tested this.

### DIFF
--- a/index/extract_refs.go
+++ b/index/extract_refs.go
@@ -640,7 +640,16 @@ func (index *SpecIndex) ExtractComponentsFromRefs(refs []*Reference) []*Referenc
 			index.refLock.Unlock()
 		} else {
 			index.refLock.Unlock()
+			// If it's local, this is safe to do unlocked
+			uri := strings.Split(ref.FullDefinition, "#/")
+			unsafeAsync := len(uri) == 2 && len(uri[0]) > 0
+			if unsafeAsync {
+				index.refLock.Lock()
+			}
 			located := index.FindComponent(ref.FullDefinition)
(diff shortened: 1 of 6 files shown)


TristanSpeakEasy:
@ThomasRooney can you add a test to validate the correct behaviour

daveshanley (approved):
I am going to test this locally first before I approve, because it looks solid. 

If you have fixed this bug, I will owe you a bottle of whiskey.

daveshanley:
I tested it, seems to work fine - however the bug is still there, when setting `ExtractRefsSequentially: false` there are still a bunch of un-bundled references. 

So, the hunt still continues, but this LGTM.

PR #339, opened 2024-10-18
fix: indexing of operations which are `$ref`s

mfbx9da4:
**Why**
We have a linting rule which checks for duplicate operation IDs / duplicate operations. This linting rule wasn't catching duplicates if the operations were living in an external $ref.

**What**
 Currently when building the index of operations we weren't checking for refs, this PR changes that.

### DIFF
--- a/index/rolodex_test.go
+++ b/index/rolodex_test.go
@@ -4,8 +4,6 @@
 package index
 
 import (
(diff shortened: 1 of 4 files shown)


daveshanley (approved):
This looks good to me! thank you for your contribution.

PR #342, opened 2024-10-22
Added a high level cache.

daveshanley:
high memory use is a problem in libopenapi, this change moves us a step closer to fixing that. Very reference heavy specs can create an ungodly number of schema proxy calls when the model is being walked.

For example the docusign spec (which has been added to the test suite) creates 4m proxy calls. This is not required as it’s the same ref being called over and over and we already have it after it’s been done once. This change adds a high level cache to prevent the over production of schemas.

### DIFF
--- a/index/cache.go
+++ b/index/cache.go
@@ -3,38 +3,136 @@
 
 package index
 
-import "sync"
+import (
+	"sync"
+	"sync/atomic"
+)
 
+// SetCache sets a sync map as a temporary cache for the index.
+// TODO: this needs to be moved to the Cache interface.
 func (index *SpecIndex) SetCache(sync *sync.Map) {
 	index.cache = sync
 }
 
 // HighCacheHit increments the counter of high cache hits by one, and returns the current value of hits.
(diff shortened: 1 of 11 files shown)


index/cache.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (index *SpecIndex) SetCache(sync *sync.Map) {
+	index.cache = sync
+}
+
+// HighCacheHit increments the counter of high cache hits by one, and returns the current value of hits.
+func (index *SpecIndex) HighCacheHit() uint64 {
+	index.highModelCacheHits.Add(1)
+	return index.highModelCacheHits.Load()
+}
```
lobocv, line 16: I would suggest you create a custom type that is a wrapper around the sync.Map and expose setter/getter methods. This way the caller doesn't have to remember to call the CacheHit/Miss functions and your metrics will be much more reliable.

lobocv (commented):
Looks good. I left some recommendations on the cache implementation but overall 👍🏽

PR #343, opened 2024-10-24
Add `nil` check on Document Render to avoid panic

inFocus7:
- Expanded the existing check during `Render` to generally check for a `nil` `highOpenAPI3Model`, with more specific error message depending on if it's Swagger or not.
- Also added a check for the `info` which is a pointer, so could _technically_ be `nil`.
- Added a test case to ensure we error. Without my changes the test panics which we're trying to avoid.

Context: I've previously hit a `panic` while learning about & implementing the library. It took me a while to figure out why and got me a bit spooked to have this in a long-running cluster/container. I added a `recover` in my end, but handling it here cleanly is nicer.

### DIFF
--- a/document_test.go
+++ b/document_test.go
@@ -427,7 +427,7 @@ func TestDocument_Render(t *testing.T) {
 		h.Components.SecuritySchemes.GetOrZero("petstore_auth").Flows.Implicit.AuthorizationUrl)
 }
 
-func TestDocument_Render_WithError(t *testing.T) {
+func TestDocument_Render_Missing_Model_Error(t *testing.T) {
 	// load an OpenAPI 3 specification from bytes
 	petstore, _ := os.ReadFile("test_specs/petstorev3.json")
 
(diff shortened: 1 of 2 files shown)


inFocus7:
Looks like its not full coverage, ill fix that in the morning :thumbsup:

document_test.go
```diff
@@ -427,6 +427,32 @@ func TestDocument_Render(t *testing.T) {
...
+	assert.Equal(t, "unable to render, no openapi model has been built for the document", e.Error())
+}
+
+func TestDocument_Render_Missing_Info_Error(t *testing.T) {
+	doc := &document{
+		// set the highOpenAPI3Model to a non-nil model to mock an existing model
+		highOpenAPI3Model: &DocumentModel[v3high.Document]{},
+		// do not set the info property
+		info: nil,
+	}
```
inFocus7, line 450: Not sure if there's a better way to set this up. I was going to try create a new document from petstore and setting its `Info` to nil, but I didn't see a way to do so.

daveshanley: This seems fine to me.

daveshanley (approved):
LGTM! Thank you for your contribution.

Issue #333, opened 2024-09-30
Data Race using GetAllComponentSchemas


califlower:
```bash
WARNING: DATA RACE
Read at 0x00c0001f82d0 by goroutine 104:
  github.com/pb33f/libopenapi/index.(*SpecIndex).GetAllComponentSchemas()
      /Users/<hidden>/Documents/edge-projects/rest-proxy/vendor/github.com/pb33f/libopenapi/index/spec_index.go:331 +0x38
  github.com/pb33f/libopenapi/index.(*SpecIndex).GetAllComponentSchemas-fm()
      <autogenerated>:1 +0x34
  github.com/pb33f/libopenapi/datamodel/low.LocateRefNodeWithContext()
      /Users/<hidden>/Documents/edge-projects/rest-proxy/vendor/github.com/pb33f/libopenapi/datamodel/low/extraction_functions.go:88 +0x434
  github.com/pb33f/libopenapi/datamodel/low/base.ExtractSchema()
      /Users/<hidden>/Documents/edge-projects/rest-proxy/vendor/github.com/pb33f/libopenapi/datamodel/low/base/schema.go:1287 +0x320
  github.com/pb33f/libopenapi/datamodel/low/v3.(*MediaType).Build()
      /Users/<hidden>/Documents/edge-projects/rest-proxy/vendor/github.com/pb33f/libopenapi/datamodel/low/v3/media_type.go:95 +0x464
  github.com/pb33f/libopenapi/datamodel/low.ExtractMapExtensions[go.shape.*github.com/pb33f/libopenapi/datamodel/low/v3.MediaType,go.shape.e114033d8920fd47cda95878864a7bac19629718fa4165c8b0cecc0ea9ec74d0].func3()
      /Users/<hidden>/Documents/edge-projects/rest-proxy/vendor/github.com/pb33f/libopenapi/datamodel/low/extraction_functions.go:760 +0x200
[...]

PR #344, opened 2024-10-24
Add RW Lock to allComponentSchemas to Prevent Data Race Error

califlower:
- solves issue #333 
- Accessing GetAllComponentSchemas causes a race condition error during usage. 
- Adds a RWLock to ensure that the map object is safely created across go routines. This pattern is used for other similar fields

### DIFF
--- a/index/index_model.go
+++ b/index/index_model.go
@@ -254,7 +254,8 @@ type SpecIndex struct {
 	allComponentSchemaDefinitions       *sync.Map                                     // all schemas found in components (openapi) or definitions (swagger).
 	securitySchemesNode                 *yaml.Node                                    // components/securitySchemes node
 	allSecuritySchemes                  map[string]*Reference                         // all security schemes / definitions.
-	allComponentSchemas                 map[string]*Reference                         // all component schemas
+	allComponentSchemas                 map[string]*Reference                         // all component schema definitions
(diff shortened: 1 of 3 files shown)


califlower:
I can add a test here, just wasn't sure about testing the race condition it was kind of difficult

califlower:
I added a concurrent access test and a nil test. That should keep the coverage where it was at before

califlower:
RIP. Doesn't seem like codecov is correct :<

daveshanley:
The codecov is correct, I checked out your branch and this is what happens after running all tests. 

Screenshot 2024-10-29 at 10 30 30 AM

The problem is the test. 

```go
func TestSpecIndex_GetAllComponentSchemas_NilIndex(t *testing.T) {
	index := &SpecIndex{}
	schemas := index.GetAllComponentSchemas()
	assert.Nil(t, schemas, "Expected GetAllComponentSchemas to return nil when index is nil")
}```

`index` cannot be nil here, it will never be nil. To fix it: 

```go
func TestSpecIndex_GetAllComponentSchemas_NilIndex(t *testing.T) {
	var index *SpecIndex
	schemas := index.GetAllComponentSchemas()
	assert.Nil(t, schemas, "Expected GetAllComponentSchemas to return nil when index is nil")
}```

daveshanley (approved):

califlower:
Ah my bad. Thank you for fixing it and merging! Really appreciate it!

PR #350, opened 2024-11-11
Add support for different bundling modes

felixjung:
The current bundling approach in libopenapi automatically inlines all refs in files other than the root file. The `BundleInlineRefs` option in `datamodel.DocumentConfiguration` appears to only affect the root file. When set to `true`, refs in the root file are inlined. When set to `false`, refs within the root file, i.e. to the `components` section, are kept where they are, while refs made to other files are inlined nevertheless.

The goal of this PR is to introduce different bundling modes.

* The first mode inlines all references.
* The second mode tracks all references across all files and ensures that they are moved to the appropriate sub-section of the `components` section in the bundled spec.

The bundle mode can be configured in a new `BundlerOptions` struct passed to the bundling functions of the bundler package.


bundler/bundler.go
```diff
@@ -51,48 +66,86 @@ func BundleBytes(bytes []byte, configuration *datamodel.DocumentConfiguration) (
...
+
+		childRefs := targetIndex.ExtractRefs(refTarget.Node, refTarget.ParentNode, make([]string, 0), 0, false, "")
+
+		for _, childRef := range childRefs {
+			childRefTarget := targetMappedReferences[childRef.FullDefinition]
+			childComponents, err := bundleRefTarget(childRef, childRefTarget, opts)
+			if err != nil {
+				return nil, err
+			}
+		}
```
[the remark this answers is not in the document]
daveshanley: There are two ways to do this. 

1. Go down low **(the hard, long, but the right way)**. 

This would be to do what you said, to manipulate the model at the low end and build up the document manually by creating high level objects from low level ones, and then adding them to the correct component location.

2. Build high  **(the simpler, faster way, questionable on correctness)**

Just build out the new components using the high level, add them as references (https://pb33f.io/libopenapi/modifying/#using-references) and then render out the doc. 

Why is this questionable? well, it means if we provide this as an API, the only way to populate the low level
is to `RenderAndReload` and it would push another run of the model again. Will it cost anything, some CPU and memory - but not much to be honest. Is it right? not exactly - but will work as long as the use is documented.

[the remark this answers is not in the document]
daveshanley: So the answer to your question is, choose the path you can invest in. If you have the time and the stomach, number 1. It will make for a better API, design and a less costly implementation. 

If you do not have time to spare (like any of us) then I suggestion option 2. It will work, it's not ideal, but it's also not wrong either - just not crisp and not performant.

califlower:
If this gets done this would be amazing. I think this is what redocly-cli does. Basically preserves the refs as local refs but moves them to their correct location in a single file. Bonus is that it'll work correctly for recursive items. 

Even better if there's support for unique naming of refs as well

felixjung:
Had very little time to work on this lately, but still want to do it.

JemDay:
FYI ... I have a PR open in the libvalidator project where I was 'nudged' to follow the "Options pattern" model which enables a more fluent style to configuration options.  My initial submission followed a very similar "options model" to the one you've got here ;-)

[PR Ref](https://github.com/pb33f/libopenapi-validator/pull/112)

ccoVeille:
I love when ideas converge!

felixjung:
@JemDay thanks. Yeah, familiar with that.

bundler/bundler.go
```diff
@@ -50,49 +73,144 @@ func BundleBytes(bytes []byte, configuration *datamodel.DocumentConfiguration) (
...
+		switch opts.RelativeRefHandling {
+		case RefHandlingInline:
+			// Just deal with simple inlining.
+			sequenced.Node.Content = mappedReference.Node.Content
+		case RefHandlingCompose:
+			// Recursively collect all reference targets to be bundled into the root
+			// file.
+			bundledComponents := make(map[string]*index.ReferenceNode)
+			if err := bundleRefTarget(sequenced, mappedReference, bundledComponents, opts); err != nil {
+				return nil, err
```
felixjung, line 107: Made a tiny bit of progress today. The idea is that `bundledComponents` now contains all the `yaml.Node`s which are to be bundled into the root spec, using their `Definition` (i.e. `#/components/schemas/MySchema`) as keys.

What I'm stuck with now, is that I don't know how to actually use these nodes. I figured the "right" way, you mentioned in your previous comments, would be to manipulate the `low` model by adding to the components there. I just can't make sense of the types used there.  See my next comment.

bundler/bundler.go
```diff
@@ -50,49 +73,144 @@ func BundleBytes(bytes []byte, configuration *datamodel.DocumentConfiguration) (
...
+			}
+			value := low.ValueReference[*base.SchemaProxy]{
+				Reference: low.Reference{},
+				Value: &base.SchemaProxy{
+					Reference: low.Reference{},
+					NodeMap:   &low.NodeMap{Nodes: &sync.Map{}},
+				},
+				ValueNode: &yaml.Node{},
 			}
+			components.Value.Schemas.Value.Set(key, value)
```
felixjung, line 152: It would appear, I need to put together the `key` and `value` so I can set them in the ordered map for Schemas. However, with all the `Value`, `ValueNode`, `Reference` etc. fields, I really don't understand what I'd be doing here. 😅 Can you once again point me in the right direction, please?

bundler/bundler.go
```diff
@@ -50,49 +73,144 @@ func BundleBytes(bytes []byte, configuration *datamodel.DocumentConfiguration) (
...
-			if mappedReference != nil && mappedReference.Circular {
-				if idx.GetLogger() != nil {
-					idx.GetLogger().Warn("[bundler] skipping circular reference",
-						"ref", sequenced.FullDefinition)
-				}
+		switch defParts[2] {
+		case "schemas":
+			key := low.KeyReference[string]{
+				Value: defParts[3],
+				KeyNode: &yaml.Node{
```
daveshanley, line 137: If you're creating a `yaml.Node` pointer, you will need to make sure the `Line` and Column` are set as well.

bundler/bundler.go
```diff
@@ -50,49 +73,144 @@ func BundleBytes(bytes []byte, configuration *datamodel.DocumentConfiguration) (
...
+					Kind:  yaml.ScalarNode,
+					Style: yaml.TaggedStyle,
+					Tag:   "!!str",
+					Value: defParts[3],
+				},
+			}
+			value := low.ValueReference[*base.SchemaProxy]{
+				Reference: low.Reference{},
+				Value: &base.SchemaProxy{
```
daveshanley, line 146: Instead of creating a `SchemaProxy` like this, use the technique here: 

https://github.com/pb33f/libopenapi/blob/main/datamodel/low/base/schema_proxy_test.go#L29

For example:

```go
var sch SchemaProxy
var myNode yaml.Node
myIndex := // some index some somewhere.

// .. do stuff to the node

ctx := context.WithValue(context.Background(), "key", "value")
err := sch.Build(ctx, &myNode, myNode.Content[0], myIndex)

Now you have a ready to go low level `SchemaProxy`

To do build a high level one from it?

https://github.com/pb33f/libopenapi/blob/main/datamodel/high/base/schema_proxy_test.go#L56C2-L56C31

bundler/bundler.go
daveshanley: The `ValueNode` is the `yaml.Node` that is a pointer to the actual node in memory, the `Reference`is always there in a low model, but the properties are only filled in if the schema proxy is a reference or not.

bundler/bundler.go
felixjung: Oh no. Thanks. I thought I read somewhere that these would be set once you decide to render. This seems rather difficult.

bundler/bundler.go
felixjung: Nice, I'll check that out.

stcalica:
continuing on the work here in case @felixjung  doesn't return to it
https://github.com/pb33f/libopenapi/pull/401

daveshanley:
Hi folks, I decided to pick this up and build this feature. 

It's a work in progress, but will be ready very soon, it's 85% done. 

You can see the code here: https://github.com/pb33f/libopenapi/blob/v0.21.13/bundler/bundler_composer.go ( will be deleted when merged)

using the following sample: https://github.com/pb33f/libopenapi/blob/v0.21.13/bundler/test/specs/main.yaml

And the resulting output https://github.com/pb33f/libopenapi/blob/v0.21.13/bundler/test/specs/bundled.yaml

felixjung:
Oh that's great news! Thanks!

Issue #355, opened 2024-11-23
Invalid JSON/YAML gets ignored when a document is opened


inFocus7:
# Issue

When opening a document with invalid spec, during the decoding of a document _regardless_ of `byass`ing the document check, we ignore any errors that come up from decoding (ex. `_ = parsedNode.Decode(&jsonSpec)`). [parsing code re-used](https://github.com/inFocus7/libopenapi/blob/70f406b6cff5d98756da54c9ff8d0fa2eb6b5537/datamodel/spec_info.go#L91-L103)

## Expected Behavior

If we set `bypass=false` (or some new configuration, if preferred), we should not ignore any errors that happen during decoding.

## Reproduction

I linked a branch in **Extra Information**, but the example spec used for this is:
```yaml
openapi: "3.0.1"
info:
  version: 1.0.0
  description: This is a sample server Petstore server.
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            maximum: 100
            format: int32
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pets"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
    post:
      summary: Create a pet
      operationId: createPets
      tags:
        - pets
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
        required: true
      responses:
        '201':
          description: Null response
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
    /pets/{petId}:
    get:
      summary: Info for a specific pet
      operationId: showPetById
      tags:
        - pets
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
[...]

PR #356, opened 2024-11-25
fix: Add error handling when parsing JSON/YAML spec

inFocus7:
Resolves: https://github.com/pb33f/libopenapi/issues/355

# Context

When testing a YAML setup similar to the one added in the test, I noticed no errors were reported and any invalid spec was ignored. While it's not a breaking behavior, it could lead to missing configuration without noticing.

# Fix

To fix this we simply should _not_ ignore any errors when parsing.

# Open Questions (from me)

- Does `bypass` make sense to use as the gate for ignoring (or not) the errors?
  - Should we _never_ ignore these errors?
  - Or should we add a new field specifically for this to lessen possibility of breaking users existing environments?
      - Adding this as a possible path forward as the unit tests fail when not ignoring errors, so it may make sense to have another parameter/option.

# Concerns (from me)

- This could technically be breaking. Users with invalid configurations - those caught by the Decode & Marshal methods - would now receive errors.
  - This breakage may be avoidable if instead of returning the error when caught, we store/save it until the final `return specInfo, nil`, where we'd return any error.

### DIFF
--- a/datamodel/spec_info.go
+++ b/datamodel/spec_info.go
@@ -90,28 +90,23 @@ func ExtractSpecInfoWithDocumentCheck(spec []byte, bypass bool) (*SpecInfo, erro
 
 	parseJSON := func(bytes []byte, spec *SpecInfo, parsedNode *yaml.Node) error {
 		var jsonSpec map[string]interface{}
-		var errs []error
+		var parseErr error
+
 		if utils.IsYAML(string(bytes)) {
-			err := parsedNode.Decode(&jsonSpec)
-			if err != nil {
-				errs = append(errs, err)
-			}
-			b, err := json.Marshal(&jsonSpec)
-			if err != nil {
-				errs = append(errs, err)
-			}
+			parseErr = parsedNode.Decode(&jsonSpec)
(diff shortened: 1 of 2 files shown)


inFocus7:
interesting:
```
--- FAIL: TestDocument_Paths_As_Array (0.00s)
panic: cannot create new document: &{%!e(string=invalid character ':' after array element) %!e(int64=53)} [recovered]
```

i'm guessing this is due to an error occurring when parsing the json as it is not valid (at least according to the library). So the paths forward I see are either:
1. Live with not catching errors during parsing (essentially removing the need for this issue + pr)
2. Use another field similar to bypass for these errors, defaulting to false so it's not breaking, and noting the limitations of it.
    - I'm not certain if or where in json `[]map` is used, and if it's a special case we support, or if it's valid json but the library doesn't know it.

daveshanley:
Hi, Are you planning on completing this? or should I consider it abandoned?

daveshanley:
Wrong ticket.

daveshanley:
abandoned.

PR #364, opened 2025-01-15
Fix enum comparison by looking on the definitive value (without style)

LasneF:
Fix this issue 
https://github.com/pb33f/openapi-changes/issues/135

### DIFF
--- a/what-changed/model/schema.go
+++ b/what-changed/model/schema.go
@@ -926,10 +926,10 @@ func checkSchemaPropertyChanges(
 	j = make(map[string]int)
 	k = make(map[string]int)
 	for i := range lSchema.Enum.Value {
-		j[toString(lSchema.Enum.Value[i].Value)] = i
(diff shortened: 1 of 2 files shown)


LasneF:
Notice this is just for enum 

there is also something there

but also there 

\github\libopenapi\what-changed\model\comparison_functions.go
func CheckForModification[T any](l, r *yaml.Node, label string, changes *[]*Change, breaking bool, orig, new T) {
	if l != nil && l.Value != "" && r != nil && r.Value != "" && (r.Value != l.Value || r.Tag != l.Tag) {


isssue is that on a yaml  
 toto : "a"  
toto : a 


image

has same Value but not same tag ( one is !!int , the other is !!str) 
and so it spot a diff breaking as of today , to me the tooling is about semantic diff not yaml diff, or space diff

daveshanley (approved):
Nice fix! thank. you.

PR #368, opened 2025-01-24
Added back support for the ~ operator which finds key names

lobocv:
Bump jsonpath

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -44,9 +44,12 @@ func FindNodes(yamlData []byte, jsonPath string) ([]*yaml.Node, error) {
 	jsonPath = FixContext(jsonPath)
 
 	var node yaml.Node
-	yaml.Unmarshal(yamlData, &node)
+	err := yaml.Unmarshal(yamlData, &node)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


utils/utils.go
```diff
@@ -114,7 +115,7 @@ func FindNodesWithoutDeserializing(node *yaml.Node, jsonPath string) ([]*yaml.No
 func FindNodesWithoutDeserializingWithTimeout(node *yaml.Node, jsonPath string, timeout time.Duration) ([]*yaml.Node, error) {
 	jsonPath = FixContext(jsonPath)
 
-	path, err := jsonpath.NewPath(jsonPath)
+	path, err := jsonpath.NewPath(jsonPath, jsonpathconfig.WithPropertyNameExtension())
```
daveshanley, line 118: I know there are more instances of this in the lib, would you be able to make sure we use this everywhere?

lobocv: Ah yes you are right. I think I got them all, other than the tests, which don't seem to need it anyway

daveshanley (approved):
LGTM

Issue #376, opened 2025-02-14
`RenderAndReload()` errors out if spec has negative floating point numbers for `minimum` or `maximum`


pratik-sharma-1:
### Issue details

`RenderAndReload()` errors out if spec has negative floating point numbers for `minimum` or `maximum`
#### Spec
```
{
    "openapi": "3.0.0",
    "info": {
        "title": "dummy",
        "version": "1.0.0"
    },
    "paths": {
        "/dummy": {
            "post": {
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "type": "object",
                                "properties": {
                                    "value": {
                                        "type": "number",
                                        "format": "decimal",
                                        "multipleOf": 0.01,
                                        "minimum": -999.99
                                    }
                                }
                            }
                        }
                    }
                },
                "responses": {
                    "200": {
                        "description": "OK"
                    }
                }
            }
        }
    }
}
```

#### Sample go program
```
package main

import (
	"fmt"

	"github.com/pb33f/libopenapi"
)

const openAPISpec = `{"openapi":"3.0.0","info":{"title":"dummy","version":"1.0.0"},"paths":{"/dummy":{"post":{"requestBody":{"content":{"application/json":{"schema":{"type":"object","properties":{"value":{"type":"number","format":"decimal","multipleOf":0.01,"minimum":-999.99}}}}}},"responses":{"200":{"description":"OK"}}}}}}`

func main() {

	document, err := libopenapi.NewDocument([]byte(openAPISpec))
	if err != nil {
		panic(err)
	}

	_, errList := document.BuildV3Model()

	if len(errList) > 0 {
		fmt.Println(errList)
		panic("error encountered while building the model")
	}

	_, _, _, errList = document.RenderAndReload()
	if len(errList) > 0 {
		fmt.Println(errList)
		panic("error encountered while rendering the model")
	}
}
```

### Current behaviour
`RenderAndReload()` errors with `[yaml: cannot decode !!float -999.99 as a !!int]`

### Expected behaviour
Since `minimum` and `maximum` are defined as numbers , negative values should be valid

### Possible cause:

1.  Explicit check of value > 0 : https://github.com/pb33f/libopenapi/blob/main/datamodel/high/node_builder.go#L504-L511
2. Similar check is also present at : https://github.com/pb33f/libopenapi/blob/main/datamodel/high/node_builder.go#L495-L498

### Reference
[...]

PR #377, opened 2025-02-14
fix error on negative float and int numbers

pratik-sharma-1:
Fixes: https://github.com/pb33f/libopenapi/issues/376

Open question:

- Was the greater than zero check done to exclude the default value? I have made that assumption hence I have changed the condition to value != 0


daveshanley (approved):
This looks good to me, it's a valid use case and this solves it.

daveshanley:
Ignore the codecov drop, thats an async issue that is a github runner problem, vs your code.

Issue #383, opened 2025-02-21
Recently added HighCache contains further re-entrancy issues


MRawlings:
This is related to: #365 

Some of the issues raised there have been addressed, one remains outstanding.

The function implicated is: `GetHighCache()`

These observations are all on the file: index/cache.go; which, based on
the subset of functions we are using, is the only place _we_ are seeing
failures.

### `GetHighCache()`:

This function is not reentrant, but appears to be used as if is is.

If this function is executed concurrently in multiple threads with the same `index` it will cause a data race.

- Multiple threads can call `GetHighCache()` concurrently. 
- Each might see `index.highModelCache == nil` and execute a `CreateNewCache()` with the intention of assigning it to `index.highModelCache`.  

Depending on the scheduling, one thread may have allocated a new `highModelCache` and be doing significant work with its version of the cache, before another thread simply overwrites it with a new Cache. 
(This is the exact failure we are seeing)

This initialisation of `index.highModelCache` is not and cannot be thread-safe.

I would suggest that initialisation needs to be removed from `GetHighCache()` and added to the `NewSpecIndex` and `NewSpecIndexWithConfig` functions. 

A new function such as this would be required in `cache.go`:

```
// InitHighCache allocates a new high model cache onto the index.
func (index *SpecIndex) InitHighCache() {
	index.highModelCache = CreateNewCache()
}
```

Which could then be called from:

```
func NewSpecIndex(rootNode *yaml.Node) *SpecIndex {
	index := new(SpecIndex)
	index.InitHighCache()
	index.config = CreateOpenAPIIndexConfig()
	index.root = rootNode
	boostrapIndexCollections(index)
	return createNewIndex(rootNode, index, false)
}
```

and its partner `NewSpecIndexWithConfig`.

This would seem to guarantee that the initialisation happens just _once_ and is never needed as a side affect of the `GetHighCache()` operation. The `nil` check and initialisation should then be removed from that function.

This _will_ have an impact on multiple unit tests that use constructs such as:

```
index := &SpecIndex{}
```
or
```
index = new(SpecIndex)
```

Those rely on the allocation side effect when later calling `GetHighCache()`. These would need to call some form of `New` function to both allocate and initialise the SpecIndex with the Cache.

Something like:

```
func NewTestSpecIndex() *SpecIndex {
	index := &SpecIndex{}
	index.InitHighCache()
	return index
}
```

Tests would need to be updated to use something like:

```
index = NewTestSpecIndex()
```

[...]

PR #384, opened 2025-02-21
Fix: Address issue: #383

MRawlings:
# Context

This should address issue: #383

# Fix

Remove initialisation side effect from `GetHighCache()`

- Pull initialisation of `index.highModelCache` out of `GetHighCache()` and create a new function `InitHighCache()`
- **For active code**, call that new function everywhere a `SpecIndex` is created. This ensures that the cache *is always* initialised and no longer requires the side effect in `GetHighCache()`
- **For unit tests** that also rely on that _side-effect_.
  - Create a new unit-test helper `NewTestSpecIndex()` that initialises the high cache.
  - Use that everywhere a new `SpecIndex` is created.

### DIFF
--- a/index/cache.go
+++ b/index/cache.go
@@ -37,10 +37,12 @@ func (index *SpecIndex) GetHighCacheMisses() uint64 {
 
 // GetHighCache returns the high model cache for this index.
 func (index *SpecIndex) GetHighCache() Cache {
-	if index.highModelCache == nil {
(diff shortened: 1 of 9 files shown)


daveshanley (approved):
This looks good to me. Thank you for your contribution.

daveshanley (approved):
Looks good, thank you for taking the time to clean this up. 

I don't feel this is a breaking change, so will be a patch release.

PR #405, opened 2025-04-25
Stub at fixing leak

1000p:
This is a stub at fixing (some of) the `goroutine` leaks happening in `datamodel/v2`.
_This PR is not a complete list of where leaks are happening in the package._


daveshanley:
Swagger (v2) is no longer maintained and has been deprecated. 

https://pb33f.io/libopenapi/swagger/

PR #406, opened 2025-04-27
Bug: Render missing required property 

geniegeist:
Given a specification
```
type: [object]
required:
  - drink
  - missing
properties:
  burger:
    type: string
  fries:
    type: string
  drink:
    type: string
```
I expect a mock
```
{
  "drink": "test",
  "missing": {} // or any other value
}
```

Currently, it crashes due to dereferencing a nil pointer.

## Solution

This PR proposes rendering the missing required property by setting it to an empty object; we could also set it to any other type (like string, integers, ...). I believe the OpenAPI specification allows any type if the property is missing in the `properties` field.

## Note

I am very new to Go. Let me know if I do something very weird.


geniegeist:
Hey @daveshanley, could you take a look at this PR? I would love to hear your feedback.

PR #422, opened 2025-05-30
Fix Bundler Compose Mode Breaking with Non Filename Dots

### LINKED ISSUE #421: New Compose Mode for Bundling Fails for Refs with a "."
`a components/schemas/asdf.zxcv` will be correctly copied to the main file, but $ref referring to it will be #/components/schemas/asdf  (last dot-separated segment is removed and so the $ref points to non-existent location)

 `components/schemas/asdf-zxcv`  on the other hand works as expected

I believe I know the issue, but opening an issue so it can be tracked

I added a test (locally)

```
package bundler

import (
	"errors"
	"github.com/pb33f/libopenapi"
	"github.com/pb33f/libopenapi/orderedmap"
	"github.com/stretchr/testify/assert"
	"testing"
)

// A component key that contains a dot (“asdf.zxcv”) must *not* be shortened to
// “asdf” when we re-wire references.
func TestRenameRef_KeepsDotInComponentName(t *testing.T) {
	spec := []byte(`
openapi: 3.1.0
info:
  title: Test
  version: 0.0.0
components:
  schemas:
    "asdf.zxcv":
      type: object
    Foo:
      allOf:
        - $ref: '#/components/schemas/asdf.zxcv'
`)

	doc, err := libopenapi.NewDocument(spec)
	assert.NoError(t, err)

	v3doc, errs := doc.BuildV3Model()
	assert.NoError(t, errors.Join(errs...))

	idx := v3doc.Model.Index
	
	processed := orderedmap.New[string, *processRef]()

	got := renameRef(idx, "#/components/schemas/asdf.zxcv", processed)

	assert.Equal(t,
		"#/components/schemas/asdf.zxcv",
		got,
		"renameRef must not strip the .zxcv part from the component key")
}
```

I can submit a patch for this

califlower:
Fixes #421 

Refs with dots were failing. Fixes that behavior, while preserving the ability to actually strip the filename

### DIFF
--- a/bundler/composer_functions.go
+++ b/bundler/composer_functions.go
@@ -156,33 +156,38 @@ func remapIndex(idx *index.SpecIndex, processedNodes *orderedmap.Map[string, *pr
 
 func renameRef(idx *index.SpecIndex, def string, processedNodes *orderedmap.Map[string, *processRef]) string {
 	if strings.Contains(def, "#/") {
-
 		defSplit := strings.Split(def, "#/")
 		if len(defSplit) != 2 {
 			return def
 		}
-		split := strings.Split(defSplit[1], "/")
-		if ref := processedNodes.GetOrZero(def); ref != nil {
-			return fmt.Sprintf("#/%s/%s", strings.Join(split[:len(split)-1], "/"), ref.name)
+		ptr := defSplit[1] // e.g. components/schemas/Foo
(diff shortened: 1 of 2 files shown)


califlower:
added the test 😢

daveshanley (approved):
LGTM! Thank you.

PR #423, opened 2025-06-02
chore: exit early when failed to parse json spec

reversearrow:
This PR introduces an early return on JSON parsing failure to avoid proceeding with invalid input.


daveshanley:
This was implemented some time ago, cleaning this up

PR #432, opened 2025-06-30
Embed name of the changed parameter

sebas2day:
Just from looking at the changed parameter result array  coming from https://github.com/pb33f/libopenapi/blob/8fd18648b47e4a823edab8680e14e89feb897290/what-changed/model/operation.go#L337-L342 it's hard to understand what parameters have actually changed.

_I'm not sure if this is the way you want the model to change going forward, but having the parameter name value somewhere helps when inspecting the changes._

### DIFF
--- a/what-changed/model/parameter.go
+++ b/what-changed/model/parameter.go
@@ -17,6 +17,7 @@ import (
 // ParameterChanges represents changes found between Swagger or OpenAPI Parameter objects.
 type ParameterChanges struct {
 	*PropertyChanges
+	Name             string            `json:"name,omitempty" yaml:"name,omitempty"`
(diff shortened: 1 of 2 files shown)


daveshanley (approved):
This is great, helpful. Thank you!

daveshanley:
> Just from looking at the changed parameter result array coming from
> 
> https://github.com/pb33f/libopenapi/blob/8fd18648b47e4a823edab8680e14e89feb897290/what-changed/model/operation.go#L337-L342
> 
> it's hard to understand what parameters have actually changed.
> _I'm not sure if this is the way you want the model to change going forward, but having the parameter name value somewhere helps when inspecting the changes._

No, this is great. Thanks for the upgrade!

Issue #433, opened 2025-07-09
Discriminator Mappings Not Updated During Bundling


califlower:
## Problem 1 Standard Bundling
Standard Bundler Inlines oneOf references that contain a discriminator mapping. `oneOf` or `anyOf` with a discriminator. Discriminators cannot be inlined as per https://redocly.com/learn/openapi/discriminator#discriminator-is-described-inline

## Problem 2 Compose Bundling
The Compose Bundler will move discriminator mapping referenced schemas under the shared schemas, but will not update the discriminator mapping

## Root Cause
1. Discriminator mappings are unfortunately not refs but have to be updated like refs in some situations
2. `oneOf`/`anyOf` with discriminators should not be inlined

## Current Behavior
- **Inline bundling**: `oneOf` and `anyOf` with refs + discriminator mappings are inlined, breaking the mappings
- **Composed bundling**: `oneOf` refs are moved to the shared schema in the root file without updating the mapping
## Expected Behavior
- **Inline bundling**:  `oneOf` and `anyOf` with refs + discriminators mappings should not be inlined, similar to circular refs
- **Composed bundling**: Discriminator mappings should be updated to point to the component locations where external schemas are composed (e.g., `#/components/schemas/`)

califlower:
Actually, I'll likely close this since I'm not sure I fully understand discriminator behavior. I think theoretically they should not be inlined if a ref is present?

califlower:
There's a chance I may be failing to understand something, it seems like maybe the compose mode is just broken idk

daveshanley:
I have not looked through this PR yet, but I am curious, why not just use a $ref? Every other file reference requires it, why do something different for the discriminator?

califlower:
I spent a lot of hours digging into this and it's definitely a bug but I've closed the PR for now until I can come up with a simpler solution. I think it's probably worth splitting this into two PRs for simplicity. I've also updated the issue to make it more clear, I can add exact examples in but didn't want to blow it up the issue size. 

Essentially, discriminator mappings are not refs in the traditional sense. I think the reasoning being is that they're meant to point to a schema location, but are not meant to resolve in the traditional sense. They're more of an identifier, hence why they're just strings. Like if we expanded the ref that the discriminator mapping is using, it wouldn't make sense.



@daveshanley

califlower:
PR 1 to fix the inline bundling. It should be good to go. The tests are much more comprehensive than the last PR I had. It should tackle most of the edge cases of this. 

[...]

PR #434, opened 2025-07-09
Fix discriminator mapping bundling

califlower:
Fixes bug with #433 

this would preserve discriminator schemas similarly to circular refs or put them under components/schemas and update the value (for compose) provided it's pointing to somewhere valid. Or it would just leave it alone if it's invalid

Allow me to add 500 lines of code to fix discriminators because they don't use standard ref behavior. Would love to know the historical reason for why they aren't standard refs in the open-api spec :<


califlower:
As this is, I don't think this encompasses the full spectrum of discriminator nonsense

PR #436, opened 2025-07-11
Fix Discriminator with Inline Bundling

califlower:
I've closed my previous PR and decided to work on each issue piecemeal. 

Inline Bundling Changes
- No Longer Inlines `oneOf` or `anyOf` for any ref, if it contains a discriminator mapping


daveshanley (approved):
LGTM! Thank you for the upgrade,

PR #437, opened 2025-07-12
fix(docs): readme sample panic

marsevilspirit:
## Description

The sample code provided in the documentation panics when run with the standard petstore.yaml from the OpenAPI Specification repository. 

## The Problem

When the application is executed, it crashes with a nil pointer dereference error after successfully processing the Pet schema.

```
$ go run . 
Schema 'Pet' has 3 properties
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7dfc8e]

goroutine 1 [running]:
main.main()
	/home/mars/code/learn_everything/learn-libopenapi/main.go:27 +0x1ce
exit status 2
```

The issue can be reproduced by fetching the specification file with:
```
curl https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/_archive_/schemas/v3.0/pass/petstore.yaml > petstorev3.json
```

## Cause of the Panic

The panic occurs because the code unconditionally attempts to access the .Properties field for every schema listed under components. However, not all schemas are of type: object.

As seen in the YAML, the Pets schema is defined as type: array. Array-type schemas do not have a .Properties field; instead, they use an items field. The code's attempt to access the nil Properties field on the Pets schema is what triggers the runtime error.

```yaml
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Pets:
      type: array
      maxItems: 100
      items:
        $ref: "#/components/schemas/Pet"
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

## The Solution

The fix makes the sample code more robust by adding a check to ensure a schema's type is object before attempting to access its Properties. Schemas of other types (like array) are now gracefully skipped, preventing the panic and allowing the program to run to completion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,12 +150,9 @@ func main() {
 	}
 
 	// The following fails after the first iteration
-	for schemaPairs := docModel.Model.Components.Schemas.First(); schemaPairs != nil; schemaPairs = schemaPairs.Next() {
-		schemaName := schemaPairs.Key()
-		schema := schemaPairs.Value()
-		properties := schema.Schema().Properties
-		if properties != nil {
-			fmt.Printf("Schema '%s' has %d properties\n", schemaName, properties.Len())
+	for schemaName, schema := range docModel.Model.Components.Schemas.FromOldest() {
+		if schema.Schema().Properties != nil {
+			fmt.Printf("Schema '%s' has %d properties\n", schemaName, schema.Schema().Properties.Len())
 		}
 	}
 }


daveshanley (approved):

daveshanley:
Thanks, I actually need to update this to use the upgraded `range` over function so:

```go
for schemaPairs := docModel.Model.Components.Schemas.First(); schemaPairs != nil; schemaPairs = schemaPairs.Next() {
    schemaName := schemaPairs.Key()
    schema := schemaPairs.Value()
    fmt.Printf("Schema '%s' has %d properties\n", schemaName, schema.Schema().Properties.Len())
}
```

becomes:

```go
for schemaName, schema := range docModel.Model.Components.Schemas.FromOldest() {
    fmt.Printf("Schema '%s' has %d properties\n", schemaName, schema.Schema().Properties.Len())
}
```

You could add that in to this update also if you want, save me doing it.

PR #441, opened 2025-07-26
Add a failing test with self reference

bcomnes:
I'm working through a bug that manifested between  v0.21.12 and  v0.23.0 (noticed in openapi-changes v0.0.81).

Basically, if there is any self reference in a schema, the tool always reports no changes. Maybe self reference is bad? But this is an incorrect failure mode either way. I suspect they should work because they used to.

### DIFF
--- a/what-changed/what_changed_test.go
+++ b/what-changed/what_changed_test.go
@@ -42,6 +42,125 @@ func TestCompareSwaggerDocuments(t *testing.T) {
 	assert.Equal(t, 27, changes.TotalBreakingChanges())
 }
 
+// TestCacheCollisionSelfReference reproduces the cache collision bug with self-referencing schemas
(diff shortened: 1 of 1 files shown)


what-changed/what_changed_test.go
```diff
@@ -42,6 +42,125 @@ func TestCompareSwaggerDocuments(t *testing.T) {
...
+  "components": {
+    "schemas": {
+      "TreeNode": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "children": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/TreeNode"
```
bcomnes, line 78: The self reference here causes the diff to always report no change (despite there being obvious changes).

PR #443, opened 2025-07-30
Fix Discriminator with Composed Bundling

### LINKED ISSUE #433: Discriminator Mappings Not Updated During Bundling

califlower:
Composed Bundling Changes

- Updates refs in discriminator mappings under composed bundling

fixes #433

### DIFF
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -120,15 +120,17 @@ func compose(model *v3.Document, compositionConfig *BundleCompositionConfig) ([]
 	rolodex := model.Rolodex
 	indexes := rolodex.GetIndexes()
 
+	discriminatorMappings := collectDiscriminatorMappingNodes(rolodex)
+
 	cf := &handleIndexConfig{
-		idx:               rolodex.GetRootIndex(),
-		model:             model,
-		indexes:           indexes,
-		seen:              sync.Map{},
-		refMap:            orderedmap.New[string, *processRef](),
-		compositionConfig: compositionConfig,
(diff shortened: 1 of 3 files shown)


daveshanley:
> I think I'll consolidate the tests, not sure if we need like 7 discriminator tests, rather than 1 test that has all the cases

It does not bother me, I actually prefer individual tests, it makes debugging specific parts of units easier in my opinion.

PR #450, opened 2025-08-11
Add SetSeed method to MockGenerator and SchemaRenderer for Deterministic Output

califlower:
- Add SetSeed() method to SchemaRenderer to allow setting custom random seed
- Add SetSeed() method to MockGenerator that delegates to underlying renderer
- Replace global rand usage with instance-based *rand.Rand in SchemaRenderer
- Update all rand.* calls to use instance-specific random source

This enables deterministic mock generation for testing purposes while keeping existing random behavior by default.

Some background, the random behavior was convenient but made certain generation scenarios a bit hard to test. If this isn't something you think is needed lmk it's not a big deal either way

### DIFF
--- a/renderer/mock_generator.go
+++ b/renderer/mock_generator.go
@@ -69,6 +69,12 @@ func (mg *MockGenerator) DisableRequiredCheck() {
 	mg.renderer.DisableRequiredCheck()
 }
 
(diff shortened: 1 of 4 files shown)


califlower:
fixed by #454

daveshanley (approved):
LGTM! Thanks! this is a nice upgrade.

PR #455, opened 2025-08-20
Fix Race Condition in Doctor + Other Usages with Cache Usage

### LINKED ISSUE #453: Occasional Race Condition in Cache
I can reliably trigger a race condition with this test, but it occurs occasionally during builds.

```go
func TestSchemaProxy_ConcurrentCacheAccess(t *testing.T) {
	// Create schema that will be cached
	const ymlComponents = `components:
  schemas:
    TestSchema:
      type: object`

	var idxNode yaml.Node
	err := yaml.Unmarshal([]byte(ymlComponents), &idxNode)
	assert.NoError(t, err)
	idx := index.NewSpecIndexWithConfig(&idxNode, index.CreateOpenAPIIndexConfig())

	// Create multiple proxies that will share the same cache entry
	proxies := make([]*SchemaProxy, 10)
	for i := range proxies {
		const ymlSchema = `$ref: '#/components/schemas/TestSchema'`
		var node yaml.Node
		yaml.Unmarshal([]byte(ymlSchema), &node)
		
		lowProxy := new(lowbase.SchemaProxy)
		lowProxy.Build(context.Background(), nil, node.Content[0], idx)
		
		proxies[i] = NewSchemaProxy(&low.NodeReference[*lowbase.SchemaProxy]{
			Value: lowProxy, ValueNode: node.Content[0],
		})
	}

	// Trigger race by having all proxies access Schema() simultaneously
	var wg sync.WaitGroup
	for _, proxy := range proxies {
		wg.Add(1)
		go func(p *SchemaProxy) {
			defer wg.Done()
			schema := p.Schema() // This should trigger the race with old code
			assert.NotNil(t, schema)
		}(proxy)
	}
	wg.Wait()
}
```

This issue is here 

```go

func (sp *SchemaProxy) Schema() *Schema {
	if sp == nil || sp.lock == nil {
		return nil
	}

	sp.lock.Lock()
	defer sp.lock.Unlock()

	if sp.rendered != nil {
		return sp.rendered
	}

	if sp.schema == nil || sp.schema.Value == nil {
		return nil
	}

	// check the high-level cache first.
	idx := sp.schema.Value.GetIndex()
	if idx != nil && sp.schema.Value != nil {
		if sp.schema.Value.IsReference() && sp.schema.Value.GetReferenceNode() != nil && sp.schema.GetValueNode() != nil {
			loc := fmt.Sprintf("%s:%d:%d", idx.GetSpecAbsolutePath(), sp.schema.GetValueNode().Line, sp.schema.GetValueNode().Column)
			if seen, ok := idx.GetHighCache().Load(loc); ok {
				idx.HighCacheHit()
				// attribute the parent proxy to the cloned schema
				schema := (*seen.(*Schema))
				schema.ParentProxy = sp
				return &schema
			} else {
				idx.HighCacheMiss()
			}
		}
	}
```

```go
				schema := (*seen.(*Schema))
				schema.ParentProxy = sp
				return &schema
```

schema.ParentProxy = sp causes the race condition as far as I can tell. 

```go
schema := (*seen.(*Schema))
schema.ParentProxy = sp
```

Removing this seems to work? But not sure if it's a good idea. My assumption is that parent proxy should be set from when the cache is made anyways so it might be no issue

califlower:
fixes #453

This is also cooking doctor https://github.com/pb33f/doctor/issues/67

### DIFF
--- a/datamodel/high/base/schema_proxy.go
+++ b/datamodel/high/base/schema_proxy.go
@@ -116,8 +116,6 @@ func (sp *SchemaProxy) Schema() *Schema {
 			loc := fmt.Sprintf("%s:%d:%d", idx.GetSpecAbsolutePath(), sp.schema.GetValueNode().Line, sp.schema.GetValueNode().Column)
 			if seen, ok := idx.GetHighCache().Load(loc); ok {
 				idx.HighCacheHit()
-				// Return cached schema without modification to avoid race conditions
-				// NOTE: ParentProxy may not be set correctly for cached schemas
 				return seen.(*Schema)
 			} else {
 				idx.HighCacheMiss()
--- a/datamodel/high/base/schema_proxy.go
+++ b/datamodel/high/base/schema_proxy.go
@@ -65,6 +65,13 @@ func NewSchemaProxy(schema *low.NodeReference[*base.SchemaProxy]) *SchemaProxy {
 	return &SchemaProxy{schema: schema, lock: &sync.Mutex{}}
 }
 
+// copySchemaWithParentProxy creates a shallow copy of a schema and sets the ParentProxy
+func (sp *SchemaProxy) copySchemaWithParentProxy(schema *Schema) *Schema {
+	schemaCopy := *schema
+	schemaCopy.ParentProxy = sp
+	return &schemaCopy
+}
+
 // CreateSchemaProxy will create a new high-level SchemaProxy from a high-level Schema, this acts the same
 // as if the SchemaProxy is pre-rendered.
 func CreateSchemaProxy(schema *Schema) *SchemaProxy {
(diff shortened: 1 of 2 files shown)


daveshanley:
The doctor fix has been merged (thanks!) but we need to think about this one, we cannot undo that wiring as it's used in other applications. Undoing it makes me nervous as I don't know what the downstream effects could be, even if our tests all still pass. 

then again, maybe I am over thinking it. 

however, I would like to try the other solution you recommended - moving the wiring to cache creation time, vs on hit.

califlower:
Okay, so I couldn't create it at cache wiring time, but I managed to get it working properly with a shallow copy. I think that should preserve the intent. I also added a test to check the parent relationship so that should fail if anyone tries to remove that specific line in the future

PR #465, opened 2025-08-30
Attempting to fix #464

### LINKED ISSUE #464: Fail to resolve example if referenced 2 times
Hello,

I've attached an example with a valid spec and references. When referencing an example two times the resolver gives up and puts the full file path into the spec versus a local reference. Let me know if I can provide any more information.

main.go

```go
package main

import (
	"fmt"
	"os"
	"path/filepath"

	"github.com/pb33f/libopenapi/bundler"
	"github.com/pb33f/libopenapi/datamodel"
)

// REPLACE THESE FILEPATHS WITH RESPECT TO LOCAL MACHINE
var filePath = ".\\foobar\\v1\\specs\\foobar-api.oas3.yaml"
var outputPath = ".\\bundled.yaml"

func main() {
	output, err := BundleToBytes(filePath)
	if err != nil {
		fmt.Println(err)
		return
	}

	os.WriteFile(outputPath, output, 0644)
}

func BundleToBytes(oasFile string) ([]byte, error) {
	specBytes, err := os.ReadFile(oasFile)
	if err != nil {
		return nil, err
	}

	baseDir := filepath.Dir(oasFile)

	cfg := datamodel.DocumentConfiguration{
		BasePath:                baseDir,
		ExtractRefsSequentially: true,
		AllowFileReferences:     true,
		AllowRemoteReferences:   true,
	}

	var bcc = bundler.BundleCompositionConfig{}

	out, err := bundler.BundleBytesComposed(specBytes, &cfg, &bcc)
	if err != nil {
		return nil, err
	}
	return out, nil
}
```

Input Spec
```yaml
openapi: 3.0.1

paths:
  /foos:
    post:
      requestBody:
        $ref: "./components/requests/foo.yaml"
    put:
      requestBody:
        $ref: "./components/requests/bar.yaml"
```

Output Spec:
```yaml
openapi: 3.0.1
paths:
    /foos:
        post:
            requestBody:
                $ref: "#/components/responses/foo"
        put:
            requestBody:
                $ref: "#/components/responses/bar"
components:
    schemas:
        foo:
            type: object
            properties:
                foo:
                    type: string
            example:
                foo: "bar"
        bar:
            type: object
            properties:
                foo:
                    type: string
            example:
                # Expect this to be resolved or refeneced from components
                $ref: C:\Users\N813971\Downloads\libopenapi-issue-simple\foobar\v1\specs\components\examples\bar.yaml
    responses:
        foo:
            content:
                application/json:
                    schema:
                        type: object
                        properties:
                            foo:
                                $ref: "#/components/schemas/foo"
            description: ""
        bar:
            content:
                application/json:
                    schema:
                        type: object
                        properties:
                            bar:
                                $ref: "#/components/schemas/bar"
            description: ""
```

[libopenapi-issue-simple.zip](https://github.com/user-attachments/files/22052266/libopenapi-issue-simple.zip)

daveshanley:


daveshanley:
implemented in `v0.27.0`

PR #483, opened 2025-12-02
v0.29.0

daveshanley:
Configurable Breaking Change Rules

  New Feature

  - Configurable breaking change detection - All breaking change rules are now configurable instead of hardcoded, allowing users to customize what is considered a breaking change for their use case

  New Files

  - breaking_rules.go - Core API functions (GenerateDefaultBreakingRules(), IsBreakingChange(), BreakingAdded/Modified/Removed())
  - breaking_rules_model.go - Type definitions for all component rules (Schema, Operation, Parameter, etc.)
  - breaking_rules_config.go - Configuration management with Merge() and IsBreaking() lookup methods
  - breaking_rules_constants.go - Component and property name constants for type-safe lookups
  - breaking_rules_test.go - Comprehensive test coverage (710+ lines)

  Key Capabilities

  - Global config API - SetActiveBreakingRulesConfig() / GetActiveBreakingRulesConfig() for setting custom rules
  - Sparse overrides - Users only specify rules they want to change; defaults apply for everything else
  - Merge functionality - Custom configs merge cleanly with defaults
  - Thread-safe - Uses sync.RWMutex for concurrent access
  - Cached defaults - sync.Once pattern for efficient default rule generation

  Refactored Files (35+ model files)

  All comparison functions now use configurable lookups instead of hardcoded booleans:
  - schema.go, operation.go, parameter.go, path_item.go, paths.go
  - header.go, media_type.go, response.go, responses.go, request_body.go
  - security_scheme.go, security_requirement.go, oauth_flows.go, callback.go
  - info.go, contact.go, license.go, tag.go, external_docs.go
  - server.go, server_variable.go, link.go, encoding.go, discriminator.go, xml.go, example.go

  Backward Compatibility

  - Default behavior is unchanged - same rules as previously hardcoded
  - No breaking changes to existing API
  - Configuration is optional; if not set, defaults apply
  - Swagger 2.0 rules remain hardcoded (OpenAPI 3.x only)

  Other Changes

  - Fixed issue #484
  - Bumped go.yaml.in/yaml/v4 from 4.0.0-rc.2 to 4.0.0-rc.3
  - Added edge case tests for various properties
  - Improved path handling in what-changed

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,18 +51,11 @@ jobs:
       - name: Test
         run: go test ./...
       - name: Coverage
-        run: |
-          go get github.com/axw/gocov/gocov
-          go get github.com/AlekSi/gocov-xml
-          go install github.com/axw/gocov/gocov
-          go install github.com/AlekSi/gocov-xml
-      - run: |
-          go test -v -coverprofile cover.out ./...
-          gocov convert cover.out | gocov-xml > coverage.xml
+        run: go test -v -coverprofile=coverage.out ./...
       - uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: ./coverage.xml
+          files: ./coverage.out
           flags: unittests
           fail_ci_if_error: false
           verbose: true


daveshanley:
Changed code coverage to use go's built in tooling vs a third party lib, the coverage report does not make much sense, but this is the reason for the coverage changes.

PR #499, opened 2026-01-07
issue #500: Support references starting with numeric character

pete-woods:
Bundling does not currently support this format of reference.

Adding a test to help develop a fix.


pete-woods:
@daveshanley I added a "fix" for this to this PR, now I know the issue is reference query segments starting with a numeric character.

I'm unconvinced the way I've fixed it is the right one, though. I suspect the real fix should lie in the JSON path lib tokenisation (which I think is maybe incorrectly tokenising `#/400_xxxx` into separate tokens for `400` (INT) and `_xxxx` (STRING).

PR #505, opened 2026-01-21
Fix nested ref resolution for remote documents

andreyyudin:
Nested $ref values inside remote documents were resolved using the root spec’s index/path instead of the referenced document’s context, which breaks relative refs inside remote files.

**Changes**
- Resolve nested refs using the referenced document’s index/path context when configured.
- Cache compiled JSONPath expressions to reduce repeated parse/alloc overhead during node lookups and tighten timeout handling.


daveshanley:
Did you use AI for this? 

The fix is down in the guts of the system, which is highly complex! I am surprised at the intricacy of the fix - it smells like AI to me.... amirite?

PR #510, opened 2026-01-29
Fix Regression In Bundling Behavior for Circular Refs

### LINKED ISSUE #508: Bundler breaks external schemas with self-references
## Problem
When bundling OpenAPI specs, external schemas that contain self-references get inlined, but their internal `$ref` pointers remain unchanged, creating dangling references that cause composition to fail.

## Example

**main.yaml:**
```yaml
openapi: 3.1.0
paths:
  /tree:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: './external.yaml#/components/schemas/Tree'
```

**external.yaml:**
```yaml
components:
  schemas:
    Tree:
      type: object
      properties:
        name:
          type: string
        children:
          type: array
          items:
            $ref: '#/components/schemas/Tree'  # Self-reference
```

## Expected Behavior
The `Tree` schema should be copied to the root document's components with the self-reference remaining valid (or it should do what was done before, and should be left alone):
```yaml
components:
  schemas:
    Tree:
      type: object
      properties:
        name:
          type: string
        children:
          type: array
          items:
            $ref: '#/components/schemas/Tree'  # Still valid
```

## Actual Behavior
The `Tree` schema body is inlined directly into the response, but the self-reference still points to `#/components/schemas/Tree`, which doesn't exist in the root document:
```yaml
paths:
  /tree:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  name:
                    type: string
                  children:
                    type: array
                    items:
                      $ref: '#/components/schemas/Tree' 
```

Calling `BundleBytesComposed()` fails with:
```
component `#/components/schemas/Tree` does not exist in the specification
```

califlower:
Fixes a regression in the bundling behavior for circular refs with inline bundling where it wouldn't properly skip over recursive refs. Brings it closer to the original behavior. 

More details in #508 

Fixes #508

### DIFF
--- a/datamodel/high/v3/paths.go
+++ b/datamodel/high/v3/paths.go
@@ -201,9 +201,6 @@ func (p *Paths) MarshalYAMLInline() (interface{}, error) {
 			if err != nil {
 				return nil, fmt.Errorf("failed to render path '%s' inline: %w", mp.path, err)
 			}
-			if rendered == nil {
-				continue
-			}
 
 			kn := utils.CreateStringNode(mp.path)
 			kn.Style = mp.style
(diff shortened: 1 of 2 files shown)


califlower:
also fixes an issue with panics in certain ref configurations

PR #587, opened 2026-06-11
fix(v3): don't inject empty `webhooks: {}` for OpenAPI 3.0 docs with a same-named scalar value

### LINKED ISSUE #586: Empty `webhooks: {}` injected when rendering OpenAPI 3.0 docs containing a same-named scalar value (regression in v0.35.1)
### Describe the bug

When rendering an **OpenAPI 3.0.x** document that does **not** declare a top-level `webhooks` key, libopenapi injects a spurious empty `webhooks: {}` into the rendered output **if the literal string `webhooks` appears anywhere else in the document as a scalar value** (for example as the value of a vendor extension like `x-foo: { service: webhooks }`, or a server URL ending in `/webhooks`).

`webhooks` is an OpenAPI **3.1+** top-level field. Emitting it for a `3.0.x` document produces output that fails strict 3.0 schema validators (e.g. Spectral's `oas3-schema`: *"Property webhooks is not expected to be here."*).

This is a regression: `v0.34.4` is fine, `v0.35.1` through `v0.37.3` are affected.

### Minimal reproduction

```go
package main

import (
	"fmt"
	"strings"

	"github.com/pb33f/libopenapi"
)

func main() {
	src := []byte(`openapi: 3.0.3
info:
  title: t
  version: "1.0.0"
x-foo:
  service: webhooks
paths:
  /a:
    get:
      responses:
        '200':
          description: OK
`)
	doc, _ := libopenapi.NewDocument(src)
	doc.BuildV3Model()
	out, _ := doc.Render()
	fmt.Println(string(out))
	fmt.Println("injected webhooks:", strings.Contains(string(out), "\nwebhooks:"))
}
```

### Expected

The rendered document is unchanged — no `webhooks` key, because the source has no top-level `webhooks` key. (`injected webhooks: false`)

### Actual (v0.35.1 – v0.37.3)

```yaml
openapi: 3.0.3
info:
  title: t
  version: "1.0.0"
x-foo:
  service: webhooks
webhooks: {}        # <-- injected, invalid in OpenAPI 3.0.x
paths:
  /a:
    get:
      responses:
        '200':
          description: OK
```
(`injected webhooks: true`)

### Root cause

Bisected to commit `cda65e2` ("more model refactoring and cleanup", first released in `v0.35.1`).

In `datamodel/low/v3/create_document.go`, `extractWebhooks` calls `low.ExtractMap[*PathItem](ctx, WebhooksLabel, root, idx)` unconditionally and then guards on `if hooksN != nil && hooksL != nil`. When there is no genuine top-level `webhooks` key, `ExtractMap` resolves the label through the index and matches a **same-named scalar value** elsewhere in the document (e.g. the `webhooks` in `service: webhooks`). That returns a non-nil empty map plus non-nil key/value nodes, so `doc.Webhooks` is populated with an empty map, which the high-level model then renders as `webhooks: {}`.

Every sibling extractor (`extractServers`, `extractTags`, `extractPaths`, `extractComponents`) instead resolves the genuine top-level key via the pre-collected `documentTopLevelNodes` and guards on `vn != nil`, so they are not affected. `extractWebhooks` is the only one that bypasses that.

### Suggested fix

Gate `extractWebhooks` on the genuine top-level node (`nodes.webhooks`), mirroring the sibling extractors. PR to follow.

### Versions

- libopenapi: `v0.35.1` … `v0.37.3` affected; `v0.34.4` and earlier unaffected
- Go: 1.25

asadtariq96:
## What

`extractWebhooks` (in `datamodel/low/v3/create_document.go`) called `low.ExtractMap[*PathItem](ctx, WebhooksLabel, root, idx)` unconditionally and then guarded on the returned key/value nodes (`if hooksN != nil && hooksL != nil`).

When a document has **no** top-level `webhooks` key, `ExtractMap` resolves the label through the index and can match a **same-named scalar value** elsewhere in the document — e.g. the `webhooks` in `x-foo: { service: webhooks }`, or a server URL ending in `/webhooks`. That returns a non-nil empty map plus non-nil key/value nodes, so `doc.Webhooks` gets populated with an empty map, which the high-level model renders as `webhooks: {}`.

`webhooks` is an OpenAPI 3.1+ field, so emitting it into a rendered **3.0.x** document produces output that fails strict 3.0 schema validation (e.g. Spectral `oas3-schema`).

## Fix

Gate extraction on the genuine top-level `webhooks` node (`nodes.webhooks`, already collected by `collectDocumentTopLevelNodes`), mirroring `extractServers` / `extractTags` / `extractPaths`, which all resolve the real top-level key and guard on `vn != nil`. `extractWebhooks` was the only top-level extractor bypassing that.

## Regression range

Bisected to `cda65e2` ("more model refactoring and cleanup"), first released in **v0.35.1**. `v0.34.4` and earlier are unaffected; `v0.35.1`–`v0.37.3` are affected.

## Test

Added `TestCreateDocument_WebHooks_NoFalsePositive` covering a 3.0 doc whose only occurrence of `webhooks` is a vendor-extension scalar value; asserts `doc.Webhooks` is not populated. Existing webhooks tests (`TestCreateDocument_WebHooks`, `TestCreateDocument_WebHooks_Error`, the `collectDocumentTopLevelNodes` tests) still pass, and the full `go test ./...` suite is green.

Fixes #586

### DIFF
--- a/datamodel/low/v3/create_document.go
+++ b/datamodel/low/v3/create_document.go
@@ -483,6 +483,11 @@ func extractPaths(ctx context.Context, root *yaml.Node, nodes documentTopLevelNo
 }
 
 func extractWebhooks(ctx context.Context, root *yaml.Node, nodes documentTopLevelNodes, doc *Document, idx *index.SpecIndex) error {
+	// without a genuine top-level key, ExtractMap can match a same-named scalar
+	// (e.g. "webhooks" in an extension value) and create an empty webhooks map.
+	if nodes.webhooks.value == nil {
+		return nil
+	}
 	hooks, hooksL, hooksN, err := low.ExtractMap[*PathItem](ctx, WebhooksLabel, root, idx)
 	if err != nil {
 		return err
--- a/datamodel/low/v3/create_document_test.go
+++ b/datamodel/low/v3/create_document_test.go
@@ -496,6 +496,28 @@ webhooks:
 	assert.Len(t, utils.UnwrapErrors(err), 1)
 }
 
+// a "webhooks" scalar value (here in an extension) must not create a webhooks map.
+func TestCreateDocument_WebHooks_NoFalsePositive(t *testing.T) {
+	yml := `openapi: 3.0.3
+info:
+  title: t
+  version: "1.0.0"
+x-foo:
+  service: webhooks
+paths:
+  /a:
+    get:
+      responses:
+        '200':
+          description: OK`
+
+	info, _ := datamodel.ExtractSpecInfo([]byte(yml))
+	d, err := CreateDocumentFromConfig(info, &datamodel.DocumentConfiguration{})
+	require.NoError(t, err)
+	assert.Nil(t, d.Webhooks.Value)
+	assert.Nil(t, d.Webhooks.KeyNode)
+}
+
 func TestCreateDocument_Servers(t *testing.T) {
 	initTest()
 	assert.Len(t, doc.Servers.Value, 2)


daveshanley (approved):
lgtm, thanks.

PR #594, opened 2026-06-24
fix: don't mutate shared yaml.Node when rendering schemas concurrently

### LINKED ISSUE #912: `oas-schema-check` intermittently reports a false `const` type error (data race, since v0.29.3)
Since `v0.29.3`, the `oas-schema-check` rule **intermittently** reports a false
error for a perfectly valid `const`:

```
a.yaml:27:21  ✗ error  `const` value type does not match schema type [string]  oas-schema-check
```

The schema is valid OpenAPI 3.1 — `status` is `type: string` with `const: ok`,
and `"ok"` is a string, so there is nothing to flag. The error only appears on
*some* runs of the exact same input, which makes this a concurrency data race,
not a real validation finding:

- linting the offending spec **alone** trips it rarely (~1% of runs);
- linting a second, unrelated spec **before** it in the same invocation trips it
  often (~30% of runs);
- `GOMAXPROCS=1` makes it disappear entirely;
- `v0.29.2` never reproduces it.

`v0.29.3` is the release that bumped libopenapi to `v0.38.0` and "adapted the
rules to the lazy schema cache", so a shared/unsynchronized lazy schema cache is
the most likely culprit.

## Affected versions

| Version | Result |
| --- | --- |
| `v0.29.2` | not affected |
| `v0.29.3` | affected (libopenapi v0.38.0 + lazy schema cache landed here) |
| `v0.29.4` | affected (tested) |

Reproduced on both **linux/amd64** (~30% hit rate) and **darwin/arm64**
(~1.7% hit rate — same bug, just needs more iterations to observe).

```
$ vacuum version
0.29.4
```

## Reproduction files

**`a.yaml`** — the valid spec containing the `const`:

```yaml
openapi: "3.1.0"
info:
  title: A
  version: "1.0"
  description: spec A
servers:
  - url: https://a.example.com
tags:
  - name: a
    description: tag a
paths:
  /health:
    get:
      description: health
      operationId: checkHealth
      tags: [a]
      responses:
        "200":
          description: ok
          content:
            application/json:
              schema:
                type: object
                properties:
                  status:
                    type: string
                    const: ok
                    example: ok
```

**`b.yaml`** — any second, unrelated valid spec (its content is not important;
its presence in the same invocation is what raises the race's hit rate):

```yaml
openapi: "3.1.0"
info:
  title: B
  version: "1.0"
  description: spec B
servers:
  - url: https://b.example.com
tags:
  - name: b
    description: tag b
paths:
  /x:
    get:
      description: get x
      operationId: getX
      tags: [b]
      responses:
        "200":
          description: ok
          content:
            application/json:
              schema:
                type: string
                example: hi
```

## Steps to reproduce

The race is intermittent, so loop the command and count how often the false
error appears:

```sh
fails=0
for i in $(seq 1 300); do
  out="$(vacuum lint --errors b.yaml a.yaml 2>&1 || true)"
  case "$out" in *oas-schema-check*) fails=$((fails+1)) ;; esac
done
echo "$fails/300 runs hit the false const error"
```

> Note: `vacuum` exits non-zero when it reports the error, so capture the output
> (`|| true`) and grep the captured text. Piping `vacuum ... | grep` under
> `set -o pipefail` would hide the failures behind vacuum's own exit code.

Full `reproduce.sh` that also runs the alone / `GOMAXPROCS=1` / `0.29.2`
comparisons:

```sh
#!/usr/bin/env bash
set -euo pipefail

VACUUM="${VACUUM:-vacuum}"        # set VACUUM=/path/to/vacuum to test a specific build
N="${N:-300}"                     # iterations (race is intermittent; needs a loop)
cd "$(dirname "$0")"

echo "vacuum: $($VACUUM version 2>/dev/null | head -1)"
echo "GOMAXPROCS=${GOMAXPROCS:-<default>}"
echo

count() { # files... -> prints "<fails>/<N>"
  local f=0 out
  for _ in $(seq 1 "$N"); do
    out="$("$VACUUM" lint --errors "$@" 2>&1 || true)"
    case "$out" in *oas-schema-check*) f=$((f+1)) ;; esac
  done
  echo "$f/$N"
}

echo "a.yaml alone .............. $(count a.yaml)"
echo "b.yaml a.yaml batched ..... $(count b.yaml a.yaml)"
```

Run it against the broken and the last-good build:

```sh
VACUUM=./vacuum-0.29.4 N=200 ./reproduce.sh
VACUUM=./vacuum-0.29.4 N=200 GOMAXPROCS=1 ./reproduce.sh
VACUUM=./vacuum-0.29.2 N=200 ./reproduce.sh
```

## Observed results (vacuum 0.29.4, linux/amd64)

| run | false `oas-schema-check` hits |
| --- | --- |
| `a.yaml` alone | ~2/200 (~1%) |
| `b.yaml a.yaml`, default concurrency | ~61/200 (~30%) |
| `b.yaml a.yaml`, `GOMAXPROCS=1` | 0/200 |
| `b.yaml a.yaml`, vacuum 0.29.2 | 0/200 |

A full failing run (`vacuum lint --details --errors b.yaml a.yaml`):

```
 > a.yaml
 ----------------------------------------------------------------------------------------------------------

Location      Severity   Message                                                                          Rule
────────────  ─────────  ───────────────────────────────────────────────────────────────────────────────  ────────────────
a.yaml:27:21  ✗ error     `const` value type does not match schema type [string]                           oas-schema-check

 category              ✗ errors      ▲ warnings    ● info
 Schemas               1             0             0
 total                 1             0             0

 | ✗ Failed with 1 errors, 0 warnings and 0 informs.
```

The same invocation on a passing run reports `0 errors` for `a.yaml`.

## Expected

`oas-schema-check` should pass deterministically. `const: ok` is a valid string
value for `type: string`, and the result must never depend on goroutine
scheduling or on which other files happen to share the invocation.

## Likely cause / suggested next step

`GOMAXPROCS=1` eliminating the failure strongly indicates an unsynchronized
concurrent access — almost certainly in the lazy schema cache introduced in
v0.29.3. A `-race` build of vacuum run on a single `vacuum lint b.yaml a.yaml`
should flag the racy read/write directly.

gnuletik:
Fixes https://github.com/daveshanley/vacuum/issues/912

## Summary

`NodeBuilder.AddYAMLNode` calls `(*yaml.Node).Encode(value)` where `value` is frequently a `*yaml.Node` owned by the model. `Encode` runs the value through the desolve stage, which rewrites `Tag`/`Style` **in place**, and the representer returns input `*yaml.Node`s as-is (`nodev` returns the node unchanged). So rendering a schema mutates the shared node it was built from.

When schemas are rendered concurrently — e.g. a linter running rules in parallel — that mutation races with any goroutine reading the same node. A common symptom: a string scalar's tag is briefly elided to `""` mid-render, so a concurrent reader checking `node.Tag == "!!str"` sees the wrong type and reports a bogus `const value type does not match schema type` error.

This fixes it by encoding a deep copy whenever the value is a `*yaml.Node`, so the desolve mutation lands on a throwaway copy and the model's nodes stay immutable.

## When this regressed

The shared-node read path is not new — the trigger is a change in `go.yaml.in/yaml/v4`'s `(*Node).Encode`, pulled in when libopenapi went from v0.37.x to v0.38.0.

- `go.yaml.in/yaml/v4` **rc.4** (used up to libopenapi v0.37.2): `Encode` marshalled the value to YAML bytes and re-parsed them into fresh nodes. It never touched the caller's node. There is no `desolver.go` in rc.4.
- `go.yaml.in/yaml/v4` **rc.5** (pulled in by libopenapi v0.38.0): `Encode` was rewritten into a `Represent` → `Desolve` → `Serialize` pipeline. The new `Desolve` stage elides inferable tags (`n.Tag = ""`) and normalizes quote style **on the represented graph**, and `Represent`/`nodev` returns input `*Node` values as-is (aliased). So `Encode` now mutates caller-owned nodes.

Because libopenapi hands shared model nodes to `Encode` during inline schema rendering, that new in-place mutation began landing on shared nodes and racing with concurrent readers. This is why the downstream symptom (vacuum #912) appears exactly at vacuum v0.29.3 / libopenapi v0.38.0 and not before.

Downstream confirmation in vacuum (which renders schemas concurrently across rules): `GOMAXPROCS=1` makes the false error disappear, vacuum v0.29.2 (libopenapi v0.37.2 / yaml rc.4) never reproduces it, and v0.29.3+ does.

## Changes

- `datamodel/high/node_builder.go`: add `encodeSafeValue` (deep-copies `*yaml.Node` values) and `deepCopyYAMLNode`; route both `rawNode.Encode(value)` call sites through `encodeSafeValue`.
- `datamodel/high/base/schema_proxy_test.go`: add `TestSchemaProxy_ConcurrentRender_DoesNotMutateSharedNodes`, which renders a schema concurrently while reading a captured `const` scalar's tag and asserts the tag is never mutated.
- `datamodel/high/node_builder_test.go`: add `TestEncodeSafeValue_DeepCopiesYAMLNodes`, a unit test covering the passthrough, nil, alias and content branches of the new helpers.

## Reproduction

The race detector flags it on `main`:

```
WARNING: DATA RACE
Write by goroutine N:
  go.yaml.in/yaml/v4 ...Desolver.desolveScalar   desolver.go:98   (writes n.Tag/n.Style)
  go.yaml.in/yaml/v4 ...(*Node).Encode           node.go
  libopenapi ...(*NodeBuilder).AddYAMLNode       node_builder.go:648
  libopenapi ...(*Schema).MarshalYAMLInlineWithContext
Previous read by goroutine M:
  ...reads the same node's Tag
```

## Verification

| check | result |
|---|---|
| new test, `go test -race` | passes with fix; data race + FAIL without it |
| `go test ./datamodel/high/...` | pass |
| `go test ./renderer/... ./bundler/...` | pass |

## Notes

- `deepCopyYAMLNode` adds one allocation per encoded `*yaml.Node`; negligible next to the represent/desolve/serialize/recompose round-trip `Encode` already performs, and only on the `*yaml.Node` path.
- There is also an argument that `(*yaml.Node).Encode` in `go.yaml.in/yaml/v4` should not mutate caller-owned nodes at all — rc.4 did not. That would fix every caller and restore the previous contract. This PR keeps the change inside libopenapi so it doesn't depend on an upstream yaml release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/datamodel/high/base/schema_proxy_test.go
+++ b/datamodel/high/base/schema_proxy_test.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/pb33f/libopenapi/datamodel"
@@ -2214,4 +2215,108 @@ func TestCreateSchemaProxyRefWithSchema_IsRefWithSiblings(t *testing.T) {
 
 	refWithNilSchema := CreateSchemaProxyRefWithSchema("#/components/schemas/Pet", nil)
 	assert.False(t, refWithNilSchema.isRefWithSiblings())
+}
+
+// findScalarNode returns the first scalar node in the tree whose value matches.
+func findScalarNode(n *yaml.Node, value string) *yaml.Node {
+	if n == nil {
+		return nil
+	}
+	if n.Kind == yaml.ScalarNode && n.Value == value {
+		return n
+	}
+	for _, c := range n.Content {
+		if found := findScalarNode(c, value); found != nil {
+			return found
+		}
+	}
+	return nil
+}
+
+// TestSchemaProxy_ConcurrentRender_DoesNotMutateSharedNodes is a regression test
+// for a data race in inline rendering. Rendering a schema runs the underlying
+// *yaml.Node through (*yaml.Node).Encode, whose desolve stage rewrites Tag/Style
+// in place. Because the representer aliases input nodes, rendering mutated the
+// model's shared scalar nodes: while one goroutine rendered a schema, another
+// reading the same node (e.g. a const type check) could observe the tag briefly
+// cleared to "" and report a bogus "const value type does not match" error.
+//
+// The renders below must never mutate the captured `const` scalar's tag.
+// Run under `-race` to also catch the underlying data race directly.
+func TestSchemaProxy_ConcurrentRender_DoesNotMutateSharedNodes(t *testing.T) {
+	const ymlComponents = `components:
+    schemas:
+     Status:
+       type: object
+       properties:
+         state:
+           type: string
+           const: ok`
+
+	idx := func() *index.SpecIndex {
+		var idxNode yaml.Node
+		require.NoError(t, yaml.Unmarshal([]byte(ymlComponents), &idxNode))
+		return index.NewSpecIndexWithConfig(&idxNode, index.CreateOpenAPIIndexConfig())
+	}()
+
+	const ymlSchema = `type: object
+properties:
+  state:
+    type: string
+    const: ok`
+	var node yaml.Node
+	require.NoError(t, yaml.Unmarshal([]byte(ymlSchema), &node))
+
+	// The `const: ok` scalar is a string; rendering must leave its tag intact.
+	constNode := findScalarNode(node.Content[0], "ok")
+	require.NotNil(t, constNode)
+	require.Equal(t, "!!str", constNode.Tag)
+
+	lowProxy := new(lowbase.SchemaProxy)
+	require.NoError(t, lowProxy.Build(context.Background(), nil, node.Content[0], idx))
+	highProxy := NewSchemaProxy(&low.NodeReference[*lowbase.SchemaProxy]{
+		Value:     lowProxy,
+		ValueNode: node.Content[0],
+	})
+
+	var corrupted atomic.Int64
+	stop := make(chan struct{})
+
+	var readers sync.WaitGroup
+	for i := 0; i < 4; i++ {
+		readers.Add(1)
+		go func() {
+			defer readers.Done()
+			for {
+				select {
+				case <-stop:
+					return
+				default:
+					if constNode.Tag != "!!str" {
+						corrupted.Add(1)
+					}
+				}
+			}
+		}()
+	}
+
+	var writers sync.WaitGroup
+	for i := 0; i < 50; i++ {
+		writers.Add(1)
+		go func() {
+			defer writers.Done()
+			for j := 0; j < 20; j++ {
+				_, err := highProxy.MarshalYAMLInlineWithContext(NewInlineRenderContext())
+				assert.NoError(t, err)
+			}
+		}()
+	}
+
+	writers.Wait()
+	close(stop)
+	readers.Wait()
+
+	assert.Equal(t, int64(0), corrupted.Load(),
+		"rendering mutated the shared const scalar's tag (was cleared by desolve)")
+	assert.Equal(t, "!!str", constNode.Tag, "shared const scalar tag must survive rendering")
 }
--- a/datamodel/high/node_builder.go
+++ b/datamodel/high/node_builder.go
@@ -348,6 +348,36 @@ func (n *NodeBuilder) Render() *yaml.Node {
 	return m
 }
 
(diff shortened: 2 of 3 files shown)


daveshanley (approved):
LGTM, thank you!

PR #600, opened 2026-07-02
arazzo: make the retry sleeper injectable via EngineConfig.SleepFunc

khalidDaoud:
## What

Adds an optional `SleepFunc func(ctx context.Context, d time.Duration) error` to `arazzo.EngineConfig`, used for the wait between retry attempts (a failure action's `retryAfter`). `nil` preserves the existing `sleepWithContext` behavior exactly; a returned error aborts the workflow the same way a cancelled context does today.

## Why

The engine is a great fit for embedding in durable workflow runtimes (Temporal, DBOS, Restate): its control flow is deterministic given the `Executor` responses, so a host can journal each `Execute` call and replay through the engine after a crash. The one wall-clock primitive in the loop is the retry sleep (`engine.go` → `sleepWithContext`), which such hosts must substitute with their own replay-safe durable timer — a real `time.Timer` inside a replayed workflow body blocks the scheduler (Temporal's deadlock detector) or re-sleeps on recovery.

It also lets tests exercise `retryAfter` paths with a recording fake instead of a real wait.

## Notes

- No behavior change for existing users (`nil` field ⇒ current default).
- Tests cover both the substitution (recorded duration, no real sleep) and error propagation (sleeper error fails the workflow with the cause on `WorkflowResult.Error`).

Context: we're using the arazzo engine inside a durable workflow engine ([perflow](https://github.com/twinfer/perflow)), with each `Executor.Execute` journaled as an exactly-once step — this field is the last piece needed for fully replay-safe embedding.

### DIFF
--- a/arazzo/engine.go
+++ b/arazzo/engine.go
@@ -44,6 +44,19 @@ type ExecutionResponse struct {
 // EngineConfig configures engine behavior.
 type EngineConfig struct {
 	RetainResponseBodies bool // If false, nil out response bodies after extracting outputs
+
+	// SleepFunc, when non-nil, replaces the engine's default wait between
+	// retry attempts (a failure action's retryAfter). The engine calls it
+	// with the workflow context and the retryAfter duration; returning an
+	// error aborts the workflow, exactly as a cancelled context does with
+	// the default sleeper.
+	//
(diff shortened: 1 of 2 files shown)


daveshanley (approved):
nice upgrade! thanks!

PR #602, opened 2026-07-06
fix: rewrite composed refs inside extensions

matoous:
Composed bundles could leave root-level vendor extension `$refs` pointing at external files even when the referenced component had been composed into the bundle. Such refs made the bundled document not self-contained.

Allow the final composed-ref rewrite pass to also process `$ref` values under `x-*` extension fields, and add a regression test.

Spotted on our side once we upgraded to the latest vacuum version. In our case, we use extensions similarly to the included regression tests which stopped being properly composed.

Related: https://github.com/pb33f/libopenapi/commit/8585a059dd858ecd0d81d006b187bc7fbec495a6

### DIFF
--- a/bundler/bundler_ref_rewrite_test.go
+++ b/bundler/bundler_ref_rewrite_test.go
@@ -122,6 +122,57 @@ paths:
 	assertNoFilePathRefs(t, bundled)
 }
 
+func TestBundlerComposed_RewritesRootVendorExtensionRefsToComposedComponents(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	mainSpec := `openapi: 3.1.0
+info:
+  title: Root Vendor Extension Ref Test
+  version: 1.0.0
+tags:
+  - name: things
+    x-related-schemas:
+      - $ref: './models.yaml#/components/schemas/Thing'
+paths:
+  /things:
+    get:
+      tags:
+        - things
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: './models.yaml#/components/schemas/Thing'`
+
+	modelsSpec := `components:
+  schemas:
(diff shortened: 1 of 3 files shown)


daveshanley:
Thank you for your contribution!

one case that still leaves the composed bundle non-self-contained.

The patch rewrites extension $refs when the complete target exactly matches an entry in processedNodes, but not when the ref points below a composed component. For example:

```yaml
x-related-field:
  $ref: './models.yaml#/components/schemas/Thing/properties/id'
```

If `Thing` is composed elsewhere, the bundle contains #/components/schemas/Thing, but the extension still retains the external `./models.yaml#...` reference.

Could `resolveRefToComposed` find the longest composed ancestor and append the remaining JSON Pointer suffix? A regression test using a nested target such as `Thing/properties/id` would cover this case.

matoous:
@daveshanley thanks, make sense! I tried to address it but I am not 100% confident I got it right, please let me know if additional adjustments are needed.

daveshanley:
Thanks, this addresses the nested-pointer case, and I verified the regression against current main.

I found one remaining performance concern: `composedRefFor` scans the entire `processedNodes` map for every external ref rewrite, making the final pass `O(refs × composed nodes)`.

A focused benchmark scales from ~0.40 µs at 100 nodes to ~40.7 µs at 10,000 nodes.

Could we preserve the exact `O(1)` lookup first, then find an ancestor by progressively trimming JSON Pointer segments and calling `GetOrZero`? That would retain longest-ancestor behavior while making lookup proportional to pointer depth rather than bundle size.

speed is everything. *everything*

matoous:
Haha, fair point, updated!

I see that the code coverage now dropped because we are exercising path that we previously didn't. Do you want me to address that as well?

PR #606, opened 2026-07-16
Make PathItem.Build() a bit more robust to malformed input

justinclift-prvidr:
Stumbled over a panic when running vacuum on a slightly invalid `openapi.yml` file, which had an empty `parameters:` line under an operation, with no nodes underneath.

Tracing that back, it turns out libopenapi does a two pass construction in `PathItem.Build()` when creating the model and when it hits a problem it can leave nil references in the model.

The consumer of this (doctor → vacuum in this case) may not handle a model with nil references well (ie panic).

This PR takes the fairly simple approach of making this code more robust so the construction continues to completion, returning an `errors.Join()` of all the failing operations and avoiding leaving nil references.  This should help consumers avoid these panics.

Weirdly (?), `Paths.Build()` seems to swallow errors so this problem can't be easily detected by consumers of the model.  That seems to be a deliberate design choice (with a test for it!), so this PR just works with it. :smile:

### DIFF
--- a/datamodel/low/v3/path_item.go
+++ b/datamodel/low/v3/path_item.go
@@ -432,10 +432,7 @@ func (p *PathItem) Build(ctx context.Context, keyNode, root *yaml.Node, idx *ind
 		}
 		return nil, nil
 	}
-	err := datamodel.TranslateSliceParallel[low.NodeReference[*Operation], any](ops, translateFunc, nil)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


justinclift-prvidr:
Hmmm, that looks like a false coverage failure?

justinclift-prvidr:
Running the coverage tests here, those numbers don't seem to be stable anyway.

So, it doesn't seem like this PR's codecov failure is valid. :frowning:

Issue #607, opened 2026-07-20
Composed bundler panics with "index out of range" when a $ref resolves to an odd-length array (e.g. root `tags`)


jorgembfigueira:
## Summary

`BundleBytesComposed` (composed bundling) panics with `runtime error: index out of range [N] with length N` when the document contains a `$ref` that resolves to an **array (sequence) node with an odd number of elements**. A top-level `tags: {$ref: tags.yaml#/tags}` pointing at an odd-length tag list reproduces it reliably.

- **libopenapi:** v0.37.2 (hit via `vacuum` v0.27.0 `bundle --composed`)
- **Default (inline) bundling** does not panic; only `--composed` / `BundleBytesComposed` does.

## Minimal reproduction

`tags65.yaml` (65 entries — odd):
```yaml
tags:
- name: Tag0
  description: d0
# ... 65 entries total
```

`spec65.yaml`:
```yaml
openapi: 3.0.2
info: {title: t, version: 1.0.0}
tags:
  $ref: tags65.yaml#/tags
paths:
  /x:
    get:
      responses:
        "200": {description: ok}
```

```
vacuum bundle --composed spec65.yaml out.json
```

### It depends purely on array-length parity

| `tags` length | `bundle --composed` |
|---|---|
| 64 (even) | ✅ success, bundle written |
| 65 (odd)  | ❌ panic: `index out of range [65] with length 65` |
| 63 (odd)  | ❌ panic: `index out of range [63] with length 63` |

Generate the tags file with, e.g.:
```
node -e "let s='tags:\n'; for(let i=0; i<65; i++)s+='- name: Tag'+i+'\n  description: d'+i+'\n'; require('fs').writeFileSync('tags65.yaml',s)"
```

## Panic / stack trace

```
panic: runtime error: index out of range [65] with length 65

goroutine 1 [running]:
github.com/pb33f/libopenapi/bundler.walkAndRewriteRefs(...)
	.../libopenapi@v0.37.2/bundler/composer_functions.go:735 +0x1e0
github.com/pb33f/libopenapi/bundler.walkAndRewriteRefs(...)
	.../libopenapi@v0.37.2/bundler/composer_functions.go:746 +0x1c4
github.com/pb33f/libopenapi/bundler.walkAndRewriteRefs(...)
	.../libopenapi@v0.37.2/bundler/composer_functions.go:719 +0x84
github.com/pb33f/libopenapi/bundler.rewriteAllRefs(...)
[...]

PR #608, opened 2026-07-20
fix: guard composed ref walker against odd-length mapping content (#607)

SAY-5:
`BundleBytesComposed` panics with `index out of range` when a $ref resolves to an array with an odd number of elements (for example a root-level `tags` list that arrives as a mapping-tagged node during composition). `walkAndRewriteRefs` steps through a MappingNode's content two at a time assuming key/value pairs and reads `Content[i+1]`, which is out of bounds when the length is odd. This adds a bounds check: if the final key has no paired value, the dangling node is walked directly instead of indexing past the end. Even-length arrays are unaffected. Repro and fix verified against the reproduction in #607, and the full `bundler` suite plus `go test ./...` pass.

### DIFF
--- a/bundler/bundler_issue607_test.go
+++ b/bundler/bundler_issue607_test.go
@@ -1,0 +1,54 @@
+// Copyright 2026 Princess Beef Heavy Industries / Dave Shanley
+// SPDX-License-Identifier: MIT
+
+package bundler
+
+import (
+	"fmt"
+	"os"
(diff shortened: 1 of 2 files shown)


daveshanley (approved):

SAY-5:
Anything else you want on this one before it goes in? Happy to rebase if it has drifted.

PR #609, opened 2026-07-22
tests: require opt-in for remote fixtures

HNO3Miracle:
Gate tests that fetch GitHub, DigitalOcean, or other remote specifications behind `LIBOPENAPI_RUN_NETWORK_TESTS`. The default unit suite remains fully exercised against local fixtures, while network-dependent coverage can be enabled explicitly with `LIBOPENAPI_RUN_NETWORK_TESTS=1 go test ./...`.

This makes the test suite deterministic in offline and sandboxed environments instead of relying on network availability and external repository state.

Tested with `go test ./...`.


daveshanley:
Please add

```yaml
env:
  LIBOPENAPI_RUN_NETWORK_TESTS: "1"
``` 

to the build action, otherwise you're going to break the pipeline!

PR #610, opened 2026-07-22
tests: isolate temporary module commands from GO111MODULE

HNO3Miracle:
The generator round-trip tests intentionally create temporary Go modules, but their `go run` subprocesses inherit `GO111MODULE=off` from the caller. Explicitly enable module mode for those subprocesses, alongside the existing `GOWORK=off` and `GOFLAGS=-mod=mod`, so the test behavior matches its temporary-module setup.

Testing:
- `go test ./generator/golang`
- Built the package test binary in module mode, then ran both round-trip tests with the binary environment set to `GO111MODULE=off`; both pass.


daveshanley:
this is a blast from the past - it's not been required for many years? why add it?

PR #617, opened 2026-08-14
feat(generator/golang): fold 3.1 const-based scalar oneOf/anyOf into enums

sloudel:
## Problem

OpenAPI 3.1 expresses string/number/boolean enums as `oneOf` of scalar
`const` values, e.g.

```yaml
PetStatus:
  oneOf:
    - const: available
      title: Available
    - const: pending
    - const: sold
```

The go model generator routes this through `populateUnion` and emits a
raw `XXXXUnion{Raw json.RawMessage}`; the enum values and names are lost,
and `WithEnumConstants` can generate nothing.

## Fix

In `populateUnion`, before settling `KindUnion`, fold the union into a
`KindEnum` when every non-null variant is a scalar `const` (new helper
`constScalarEnumFromVariants`). Null-valued variants become `Nullable`,
object/array/union/disambiguation variants are left as unions.

## Effects

| before | after |
|---|---|
| `XXXXUnion{Raw json.RawMessage}` | `type PetStatus string` + typed constants |

## Tests

- `TestJSONSchema202012ConstScalarEnumVariants` — const enum, string/int,
  `anyOf`, nullable, mixed->`any`, compile-check.
- `TestJSONSchemaSchemaObjectUnionNotEnum` — object variants stay unions.

Verified: `go build ./generator/...`, `go vet ./generator/golang/`,
`go test ./generator/golang/` — all green.

### DIFF
--- a/generator/golang/from_openapi.go
+++ b/generator/golang/from_openapi.go
@@ -8,6 +8,7 @@ import (
 
 	highbase "github.com/pb33f/libopenapi/datamodel/high/base"
 	"github.com/pb33f/libopenapi/orderedmap"
+	"go.yaml.in/yaml/v4"
 )
 
 func (g *Generator) irFromOpenAPI(name string, proxy *highbase.SchemaProxy, path string) (*SchemaIR, error) {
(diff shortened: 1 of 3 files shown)


generator/golang/from_openapi.go
```diff
@@ -571,6 +583,52 @@ func nonNullVariants(variants []*SchemaIR) []*SchemaIR {
...
+	nullable := false
+	for _, v := range variants {
+		if v == nil {
+			return nil, false, false
+		}
+		if isNullOnlyIR(v) {
+			nullable = true
+			continue
+		}
+		if v.Const == nil {
```
daveshanley, line 607: P1: `constScalarEnumFromVariants` must reject non-scalar `const` nodes. Untyped mapping and sequence constants pass these guards and become string enums with empty constants, so valid object/array JSON cannot unmarshal. Check `v.Const.Kind == yaml.ScalarNode` (and supported scalar tags) after null handling, with public render regressions for untyped object and array constants.

daveshanley:
Follow-up: restrict folding to scalar `const` nodes and add untyped object/array regression coverage. Once updated, rerun `go test ./generator/golang` and `go test ./...`.

PR #622, opened 2026-09-01
fix(index): keep node line index sub-linear for single-line JSON documents

### LINKED ISSUE #621: v0.38: building a single-line (minified JSON) document is O(n²) in MapNodes
## Summary

Since v0.38.0, building a document whose spec is **minified JSON (whole document on one line)** takes minutes instead of seconds. An 8 MB single-line OpenAPI 3.1 spec with ~700k YAML nodes pins one CPU core at 100% inside `index.addNodeLineEntry` and does not finish within 4+ minutes. The same file builds in ~1.5 s on v0.37.3.

## Cause

Commit 4a22282 (v0.38.0) replaced the per-line `map[int]*yaml.Node` in `index/map_index_nodes.go` with a slice per line that is scanned linearly on every insert and lookup:

```go
// nodeLineEntry is a single (column, node) pair on one line of the spec. Lines hold very
// few nodes, so a small slice scanned linearly is far cheaper than a per-line map.
```

That assumption holds for YAML but not for minified JSON, where every node has `Line == 1`. Each insert then scans every node already on that line, so `MapNodes` is O(n²) in the node count. A CPU profile of a 45 s run shows 94% of samples in `addNodeLineEntry`:

```
      flat  flat%   sum%        cum   cum%
    36.49s 93.66% 93.66%     38.20s 98.05%  github.com/pb33f/libopenapi/index.addNodeLineEntry
         0     0% 98.31%     38.21s 98.07%  github.com/pb33f/libopenapi/index.(*SpecIndex).MapNodes
         0     0% 98.31%     38.21s 98.07%  github.com/pb33f/libopenapi/index.mapNodesRecursive
```

`lookupNodeLines` has the same linear cost per `GetNode` call, which also affects `FindNodeOrigin` and the `KeyNode` lookup in `extract_refs_lookup.go`.

## Reproduction

Any large spec published as compact JSON triggers it. Cloudflare's public spec is one example (fetch it and strip newlines, or use an older minified release):

```go
data, _ := os.ReadFile("cloudflare-minified.json") // one line, ~8 MB
doc, _ := libopenapi.NewDocument(data)
model, err := doc.BuildV3Model() // never returns in a reasonable time on v0.38.x
```

A synthetic reproduction is in the PR's `BenchmarkSpecIndex_MapNodes_SingleLine` (20k schemas, ~200k nodes on one line):

| | ns/op |
|---|---|
| origin/main (v0.38.7) | 101,835,604,666 (~102 s) |
| with fix | 301,491,542 (~0.3 s) |

None of `ExtractRefsSequentially`, `SkipCircularReferenceCheck`, `SkipMetadataCollection`, or `UseSchemaQuickHash` avoid it, since `MapNodes` runs regardless.

## Environment

- libopenapi v0.38.7 (also reproduced at v0.38.0 tag range via `git tag --contains 4a22282`)
- Go 1.26, darwin/arm64

## Fix

PR incoming: append-only inserts, one stable sort per line by column when `MapNodes` finishes (collapsing duplicate columns with last-write-wins, preserving the existing parent-wins semantics), and binary-search lookups. The `[][]nodeLineEntry` type and its call sites are unchanged.

s04:
Fixes #621

## Problem

Since v0.38.0 (4a22282) the per-line node index stores entries in a slice that is scanned linearly on every insert and lookup. For YAML that is cheap, but a minified JSON document puts every node on line 1, so building the index is O(n²) in node count. An 8 MB single-line spec (~700k nodes) pins a core inside `addNodeLineEntry` for minutes; v0.37.3 builds it in ~1.5 s.

## Change

`index/map_index_nodes.go`:

- `addNodeLineEntry` is now append-only. No scan per insert.
- `mapNodesRecursive` records each node once in post-order, avoiding duplicate allocations while preserving last-write-wins behavior for parents and later siblings.
- New `sortNodeLines` runs once at the end of `MapNodes`: stable-sorts each line by column and collapses duplicate columns keeping the last write. `mapNodesRecursive` writes parents after children, so parents still win collisions, matching the previous behaviour and the original map semantics.
- `lookupNodeLines` is a binary search over the sorted line.

The `[][]nodeLineEntry` type and all call sites (`GetNode`, `GetNodeMap`, `FindNodeOrigin`, the `KeyNode` lookup in `extract_refs_lookup.go`) are unchanged. One behavioural nuance: `entries[0]` on a line is now the leftmost node rather than the first one visited. For the `KeyNode` lookup that is the intended node, and in practice they coincide because DFS visits the leftmost child first.

## Tests

- `TestSpecIndex_MapNodes_SingleLineDocument`: minified document with ~16k nodes on one line; every node is retrievable by position and the legacy map agrees with the line index.
- `TestSpecIndex_SortNodeLines_OrdersAndDedupes`: out-of-order inserts and duplicate columns.
- `TestSpecIndex_MapNodes_RecordsEachNodeOnce`: checks the existing Petstore fixture before sorting so deduplication cannot hide duplicate traversal writes.
- `TestSpecIndex_MapNodes_PostOrderCollisions`: verifies parents win over children and later siblings win over earlier siblings through `MapNodes` and `GetNode`.
- `BenchmarkSpecIndex_MapNodes_SingleLine`: ~200k nodes on one line.

```
BenchmarkSpecIndex_MapNodes_SingleLine   before (origin/main): 101,835,604,666 ns/op
                                         after:                    301,491,542 ns/op
BenchmarkSpecIndex_MapNodes              unchanged (~378 µs/op)
```

The post-order traversal keeps the ordinary-spec benchmark at ~337 KB / 4,083 allocations per operation (down from ~387 KB / 4,906 with duplicate writes). After this adjustment, the focused regression tests, single-line benchmark, and `go test ./index ./bundler ./what-changed/...` pass.

Full `go test ./...` passed for the initial change. The 8 MB real-world single-line spec builds its V3 model in ~0.7 s with this change.


index/map_index_nodes.go
```diff
@@ -124,13 +134,29 @@ func addNodeLineEntry(lines [][]nodeLineEntry, node *yaml.Node) [][]nodeLineEntr
 		copy(expanded, lines)
 		lines = expanded
 	}
-	entries := lines[line]
-	for i := range entries {
-		if int(entries[i].column) == node.Column {
-			entries[i].node = node
-			return lines
+	lines[line] = append(lines[line], nodeLineEntry{column: int32(node.Column), node: node})
```
daveshanley, line 138: The append-only change interacts with `mapNodesRecursive`, which adds each child before recursing and then adds that same child again on return. Nearly every node is therefore stored twice; `sortNodeLines` reduces the slice length but retains the oversized backing array. In the existing ordinary-spec benchmark, this moves allocation from ~337 KB/4,084 allocs to ~388 KB/4,907 allocs per operation—about 15% more bytes and 20% more objects. Could we record each node once in post-order, or compact into right-sized storage, while preserving the last-write-wins behavior?

daveshanley (commented):
Strong fix overall; one avoidable allocation-retention concern is noted inline.

index/map_index_nodes.go
s04: Thanks for the review, I've now addressed this in latest commit `a7cc4afb`. Each node is now recorded once in post-order, preserving last-write-wins behavior. Added regression tests for duplicate writes and parent/ sibling collisions. The ordinary-spec benchmark is back to ~337 KB / 4,083 allocs per operation, down from ~387 KB / 4,906.

Can remove tests if you don't find them reasonable.

PR #626, opened 2026-09-17
Add OpenAPI and Arazzo Go SDK generation

daveshanley:
OpenAPI clients currently require consumers to rebuild transport, authentication, serialization, error, and workflow behavior around the model generator. This adds a general-purpose `generator/sdk` contract layer and a resource-oriented Go emitter.

The generated Go client includes reachable models, injected HTTP execution, resolved server defaults, named OR/AND credentials with OAuth scopes, bounded responses, cancellation, request headers, typed success and declared error models, optional-nullable PATCH semantics, and finite Arazzo workflow helpers. Arazzo input schemas generate typed input models; direct step parameter and request-body bindings are validated against both the workflow and OpenAPI contracts. Unsupported wire behavior, external SDK schema references, mapped scalar parameter serialization, and unsupported workflow control flow fail generation explicitly.

Validation:

- `go test -race ./generator/sdk/... ./generator/golang ./arazzo ./arazzo/expression ./datamodel/high/v3`
- `go test ./...`
- `go vet ./generator/... ./arazzo/... ./datamodel/high/v3`
- Generated and compiled a disposable client for the `getGoals` and `getGoal` operations in `test_specs/asana.yaml`, proving default method-name collision handling on a large specification
- Generated runtime integration covers authentication alternatives, exact parameter encoding, comma-preserving simple-style path arrays, typed errors, nullable PATCH values, bounded bodies, redirect policy, concurrent use, and differently named Arazzo input-to-path bindings


generator/sdk/golang/templates/resources.tmpl
```diff
@@ -0,0 +1,132 @@
...
+{{- end}}
+		encoded, err := encodeSimple(input.{{.FieldName}}, {{.Explode}})
+		if err != nil {
+			return nil, fmt.Errorf("encode header parameter {{.Name}}: %w", err)
+		}
+		request.Header.Set({{quote .Name}}, encoded)
+	}
+{{- end}}
+{{- end}}
+	response, payload, err := resource.client.do(request, [][]securityRequirement{
```
daveshanley: This composite literal is rebuilt on every call. Escape analysis says the two slice levels stay on the stack, but every `Scopes` slice escapes to the heap because it flows into `Credential.Apply` through an interface, so each non-empty scope list is a fresh allocation per request for data that never changes. Hoist it to a package-level `var listWidgetsSecurity = [][]securityRequirement{...}` per operation, or share one variable when operations use the same table, and pass the variable. Constant data should cost nothing at request time.

For scale: one `Widgets.List` call against a stub doer is 49 allocations and 4.5KB before a single byte hits the network.

generator/sdk/golang/templates/resources.tmpl
```diff
@@ -0,0 +1,132 @@
...
+{{- range .Security}}
+		{ {{range .}}{Name: {{quote .Name}}, Scopes: []string{ {{range .Scopes}}{{quote .}}, {{end}} }}, {{end}} },
+{{- end}}
+	})
+	if err != nil {
+		return nil, err
+	}
+	if {{range $index, $status := .SuccessStatuses}}{{if $index}} || {{end}}statusMatches(response.StatusCode, {{quote $status}}){{end}} {
```
daveshanley: `statusMatches(response.StatusCode, "201")` for a three-digit literal we already have at generation time. I benchmarked it in the generated package: 45ns and one alloc per check (there is a `fmt.Sprintf("%03d")` inside), versus 0.3ns for `response.StatusCode == 201`. The template has the literal in hand, so emit the int compare for concrete codes, `response.StatusCode/100 == 2` for `2XX`, and keep the string helper only for `default`. Same fix on line 117 for the error switch and in workflows.tmpl line 46.

generator/sdk/golang/templates/client.tmpl
```diff
@@ -0,0 +1,535 @@
...
+		return false
+	}
+	if pattern[1:] == "XX" {
+		return status/100 == int(pattern[0]-'0')
+	}
+	return fmt.Sprintf("%03d", status) == pattern
+}
+
+func encodeSimple(value any, explode bool) (string, error) {
+	reflected := reflect.ValueOf(value)
```
daveshanley: We generated this code. We know the static Go type of every parameter at generation time, and yet the runtime reaches for `reflect` on every path, query and header value. `encodeSimple("acct_123")` costs 29ns and an alloc for a value that is already a string. A `*int` costs 39ns and two allocs, against 1ns for `strconv.Itoa`. Emit typed encoders (string passthrough, `strconv` for numbers and bools, a small `[]T` joiner) and drop the `reflect` import from the generated client entirely.

The `reflect.Map` branch is dead code as well. `parameterSchemaSupported` rejects object parameters at generation time, so nothing can ever reach it.

generator/sdk/golang/templates/resources.tmpl
```diff
@@ -0,0 +1,132 @@
...
+		result := &Response[{{.ResponseType}}]{StatusCode: response.StatusCode, Header: response.Header.Clone()}
+{{- if ne .ResponseType "struct{}"}}
+		if err := decodeJSON(payload, &result.Value); err != nil {
+			return nil, &ResponseDecodeError{StatusCode: response.StatusCode, Header: response.Header.Clone(), Body: append([]byte(nil), payload...), Cause: err}
```
daveshanley: `append([]byte(nil), payload...)` copies a buffer that `do()` just allocated with `io.ReadAll` and that nobody else holds. On the `APIError` path (line 113) that doubles the memory of every non-2xx response, up to the 8MB cap. Hand `payload` over directly. `response.Header.Clone()` also runs twice on the decode-error path (105 and 108); clone once.

While in `do()`: `io.ReadAll` grows from 512 bytes by doubling, so a large JSON body reallocates a dozen times on the way to the cap. Seed a `bytes.Buffer` from `Content-Length` when the server sends one.

generator/sdk/golang/templates/resources.tmpl
```diff
@@ -0,0 +1,132 @@
...
+	if resource == nil || resource.client == nil {
+		return nil, fmt.Errorf("{{.ID}}: client is required")
+	}
+{{- if .HasParameters}}
+	if input == nil {
+		input = &{{.ParamsType}}{}
+	}
+{{- end}}
+	path := {{quote .Path}}
+	query := make(url.Values)
```
daveshanley, line 52: `query := make(url.Values)` is emitted for every operation, including operations with no query parameters, and `newRequest` then calls `query.Encode()` on the empty map. Likewise, `var body any` and `contentType := ""` are emitted only to be overwritten immediately for operations with a request body. Branch in the template on `.Parameters` and `.Body` so each emitted method declares only what it uses and the generated output reads like hand-written Go.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+		responseType = selectedType
+	}
+	return responseType, successes, errorResponses, nil
+}
+
+func (e *emitter) schemaType(proxy *highbase.SchemaProxy, hint string) (string, error) {
+	if proxy == nil {
+		return "", errors.New("schema is required")
+	}
+	if proxy.IsReference() {
```
daveshanley, line 678: Two things here.

`generator/golang/names.go:178` already has `refName` doing the last-segment split. This PR exported `PublicName` for exactly this cross-package reason, so export `RefName` beside it instead of re-rolling it here and again at line 606. The `url.PathUnescape` is also the wrong escaping for a JSON pointer (that is `~1` and `~0`, and `arazzo/expression.UnescapeJSONPointer` already exists).

Second, this binds *any* `$ref` by its last segment into `components/schemas`. An external `other.yaml#/components/schemas/Account` or a `#/components/responses/Account` silently resolves to the wrong schema when the name happens to match, and fails with "not defined in components" when it does not. Check the `#/components/schemas/` prefix explicitly and send everything else through the model generator's `WithExternalRefTypeResolver` path, which is what it is for.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+	if schema == nil {
+		return "", errors.New("schema could not be resolved")
+	}
+	types := make([]string, 0, len(schema.Type))
+	for _, schemaType := range schema.Type {
+		if schemaType != "null" {
+			types = append(types, schemaType)
+		}
+	}
+	if len(types) == 1 {
```
daveshanley, line 701: This is `goType` from `generator/golang/to_go.go:378` again, minus `formatType`. The moment anyone configures a format mapping on the model generator (`date-time` to `time.Time`, say) model fields and parameter types diverge, and today there is no way to configure one at all (see my note on `Options`). One scalar mapping, owned by the model generator, consumed here.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+
+func commentLine(value string) string {
+	value = strings.TrimSpace(strings.Split(value, "\n")[0])
+	if value == "" {
+		return ""
+	}
+	return strings.ReplaceAll(value, "*/", "* /")
+}
+
+type nameRegistry struct {
```
daveshanley: Second `nameRegistry` in the generator tree. `generator/golang/name_registry.go` has one with a different collision policy (`Foo__2` there, `FooModel` then `Foo2` here). Two registries with two policies is how a nested type name and an inline schema name end up colliding on a spec we have not seen yet. Grow the existing one (a reserved-names constructor and a suffix hint are small additions) and export it, the same way `PublicName` was exported.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+			fields[inputName] = binding.Input
+			modelField := modelgen.PublicName(binding.Property)
+			view.Fields = append(view.Fields, workflowFieldView{Name: inputName, ModelType: operation.Body.Type, ModelField: modelField})
+			view.BodyFields = append(view.BodyFields, workflowAssignmentView{Target: modelField, Source: inputName})
+		}
+	}
+	return view, nil
+}
+
+func resolveWorkflowFieldTypes(source []byte, workflows []workflowView) error {
```
daveshanley: We render Go source with the model generator and then parse it back with `go/parser` to find one struct field's type, dragging `go/ast`, `go/parser` and `go/format` into the emitter for a lookup. The model generator already knows every field's Go type. It just does not expose it. Add `Fields []GeneratedField{Name, Type}` to `GeneratedType` (it already carries `Name` and `Kind`) and read that. This function goes away, along with the per-generation `types` map over every declaration and a `format.Node` per field.

generator/sdk/workflow.go
```diff
@@ -0,0 +1,218 @@
...
+	"regexp"
+	"sort"
+	"strings"
+
+	higharazzo "github.com/pb33f/libopenapi/datamodel/high/arazzo"
+	highbase "github.com/pb33f/libopenapi/datamodel/high/base"
+	"go.yaml.in/yaml/v4"
+)
+
+var statusCriterionPattern = regexp.MustCompile(`^\$statusCode\s*==\s*([1-5][0-9][0-9])$`)
```
daveshanley: `arazzo/criterion.go` already parses this grammar (`splitSimpleCondition`, then `expression.Parse` for `$statusCode`). A second regex-based parser in a sibling package means two opinions on what `$statusCode == 201` means. Export the splitter, or move it into `arazzo/expression` next to `Parse`, and use it here.

daveshanley (commented):
Good bones, wrong altitude for the hot path. The contract layer and finite Arazzo compiler have the right shape, the generated SDK compiles and passes its own `-race` suite, and determinism holds. The generated runtime still performs work at request time that was already completed at generation time. Fix that before downstream consumers take a dependency on it. The remaining findings concern reinvention and hygiene.

**What I measured** (scratchpad clone at `f614a41e`, `go vet` clean, `go test ./generator/...` green):

| Thing | Cost |
|---|---|
| `statusMatches(code, "201")` in the generated client | 45ns, 1 alloc. An int compare is 0.3ns. |
| `encodeSimple` on a `string` parameter | 29ns, 1 alloc, for a value that is already a string |
| `encodeSimple` on a `*int` parameter | 39ns, 2 allocs. `strconv.Itoa` is 1ns, 0 allocs |
| Security table literal per generated method | every `Scopes` slice heap-allocates per call for constant data |
| One generated resource call against a stub doer | 2.0µs, 4.5KB, 49 allocs before the network |
| `append([]byte(nil), payload...)` on error paths | doubles every non-2xx body, up to 8MB |
| One full `GenerateContract` on the fixture | 1.27ms, 660KB, 13.7k allocs |
| Per-file `render`: parse / execute / `format.Source` | 16µs / 16µs / ~680µs |

The emitter is fine. gofmt is 95% of its cost and that is the price of emitting Go. The generated client is where the numbers matter, and the fixes are all "emit the thing you know" instead of "look it up with reflection or `Sprintf` at runtime".

**Reinvention.** Five things this PR re-rolls that already live in the repo: `refName` (`generator/golang/names.go`), the scalar type mapping in `goType`, `nameRegistry`, `arazzo/expression.Parse` for `$inputs.x`, and `splitSimpleCondition` for `$statusCode == 201`. `PublicName` was exported for exactly this reason, so the pattern is right, it just stopped one helper early. The workflow field-type lookup parses generated Go back into an AST to find a struct field's type, which the model generator could simply expose.

**Architecture.** Two sibling packages are both named `golang`, model-generator options cannot be passed through to the SDK models, and `Generate` cannot actually be used with workflows. `Diagnostic` documents a state `Prepare` never emits. `collectSchema` walks validation-only keywords, so "reachable" over-approximates.

**Coverage chasing.** `generateContract` grew two function parameters so `coverage_test.go` can fail each stage on cue, and `shouldPointer` survives only because a test still calls it. Clean production signatures matter more than a synthetic 100% figure.

**What I checked and did not find:** `security: []` on an operation correctly opts out of document security because the high model provides an empty non-nil slice. The `**T` optional-nullable shape renders as advertised. Two generations of the same contract are byte-identical.

Twenty-five inline comments, mostly short. Order of attack: the five template comments first, then the reinvention set, then the rest.

generator/sdk/workflow.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	bindings := make([]*PayloadBinding, 0, len(payload.Content)/2)
+	for index := 0; index < len(payload.Content); index += 2 {
+		property := payload.Content[index]
+		value := payload.Content[index+1]
+		if property.Kind != yaml.ScalarNode || value.Kind != yaml.ScalarNode {
+			return nil, errors.New("request payload properties must map directly to workflow inputs")
+		}
+		const prefix = "$inputs."
+		if !strings.HasPrefix(value.Value, prefix) || len(value.Value) == len(prefix) || strings.Contains(value.Value[len(prefix):], ".") {
```
daveshanley: `arazzo/expression.Parse("$inputs.itemId")` returns `Expression{Type: Inputs, Name: "itemId"}` and rejects everything else. Use it rather than a `HasPrefix` plus `Contains(".")` reimplementation of the same rule.

generator/sdk/prepare.go
```diff
@@ -0,0 +1,264 @@
...
+	for _, parameter := range operationParameters {
+		appendOrReplace(parameter)
+	}
+	return result
+}
+
+func prepareParameter(parameter *highv3.Parameter) *Parameter {
+	required := parameter.Required != nil && *parameter.Required
+	style := parameter.Style
+	if style == "" {
```
daveshanley: Effective style and explode defaulting belongs on `highv3.Parameter`, next to `IsDefaultFormEncoding` and `IsExploded`. wiretap and vacuum need the same answer and each one currently works it out on its own. Add `EffectiveStyle()` and `EffectiveExplode()` to the high model and call them from here.

generator/sdk/prepare.go
```diff
@@ -0,0 +1,264 @@
...
+		Summary: operation.Summary, Description: operation.Description,
+		Parameters:  mergeParameters(item.Parameters, operation.Parameters),
+		RequestBody: prepareRequestBody(operation.RequestBody),
+		Responses:   prepareResponses(operation.Responses),
+	}
+	servers := operation.Servers
+	if len(servers) == 0 {
+		servers = item.Servers
+	}
+	if len(servers) == 0 {
```
daveshanley, line 119: Document servers are resolved once at line 25 and then again for every operation that inherits them (which is nearly all of them), allocating a fresh `[]string` per operation. The emitter then reads `Servers[0]` and rejects anything that differs from the document default. Resolve the document list once, share the slice, and only resolve per operation when the operation or path item actually declares its own.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+				}
+			}
+		}
+		resourceName := modelgen.PublicName(operation.Resource)
+		resource := resources[resourceName]
+		if resource == nil {
+			resource = &resourceView{Name: resourceName, TypeName: e.names.claim(resourceName+"Resource", "Resource")}
+			resources[resourceName] = resource
+			resourceMethods[resourceName] = make(map[string]string)
+			view.Resources = append(view.Resources, *resource)
```
daveshanley: `resources[resourceName] = resource` keeps a pointer to a struct that was just *copied* into `view.Resources`, so the map only ever works as an existence check, and every operation then does a linear scan of `view.Resources` (223-228) to find its resource by name. Store an index (`map[string]int`) and append straight to `view.Resources[i]`.

Line 222 then copies every `operationView` (with all its slice headers) a second time into `operationViews` so workflows can look them up by ID. A `map[string]int` pointing into the resource's `Operations` slice does the same job with no copies.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+func renderModels(emitter *emitter, packageName string) (*modelgen.GeneratedFile, error) {
+	return modelgen.NewGenerator(
+		modelgen.WithPackageName(packageName),
+		modelgen.WithGeneratedComment(true),
+		modelgen.WithOptionalNullableAsDoublePointer(true),
+		modelgen.WithTypeNameResolver(func(name string) string { return emitter.typeNames[name] }),
+	).RenderSchemas(emitter.schemas)
+}
+
+func render(name, source string, value any) ([]byte, error) {
```
daveshanley: Templates are parsed on every `render` call, three or four times per generation. Parse once at package level with `template.Must` (they are embedded constants), and `strconv.Quote` instead of `fmt.Sprintf("%q")` for `quote`.

Honest numbers so nobody over-rotates on this one. Per file on the PR's own fixture: parse is ~16µs and 197 allocs, execute is ~16µs and 118 allocs, and `format.Source` is ~680µs and ~7,000 allocs. gofmt is 95% of render, and that is the price of emitting Go. Fix the parse because it is free, and know that the real per-request cost lives in the templates, which is where my other comments are.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+
+type modelRenderer func(*emitter, string) (*modelgen.GeneratedFile, error)
+type fileRenderer func(string, string, any) ([]byte, error)
+
+// GenerateContract emits a Go SDK from a prepared contract.
+func GenerateContract(contract *sdk.Contract, options Options) (*sdk.Result, error) {
+	return generateContract(contract, options, renderModels, render)
+}
+
+func generateContract(contract *sdk.Contract, options Options, renderModel modelRenderer, renderFile fileRenderer) (*sdk.Result, error) {
```
daveshanley: `renderModel` and `renderFile` are function parameters that exist only so `coverage_test.go` can make each stage fail on cue. That is coverage chasing leaking into the production signature. The public `GenerateContract` is the seam; delete the injection and let the stage-failure tests go with it. 100% of modified lines is a number, and I would rather have 90% with a clean signature.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+	for _, child := range schema.OneOf {
+		visit(child)
+	}
+	for _, child := range schema.AnyOf {
+		visit(child)
+	}
+	for _, child := range schema.PrefixItems {
+		visit(child)
+	}
+	visit(schema.Contains)
```
daveshanley: `collectSchema` walks `contains`, `if`/`then`/`else`, `not`, `propertyNames`, `unevaluatedItems`, `contentSchema`, `dependentSchemas` and `$defs`, so any `$ref` under those pulls a component into the generated models. The model generator renders none of them into Go shape (`collectShapeDiagnostics` only reports them), which means "reachable" over-approximates here and we emit types nothing references. Walk exactly the keywords `from_openapi.go` shapes: properties, items, additionalProperties, allOf/oneOf/anyOf, patternProperties.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+			visit(child)
+		}
+	}
+}
+
+func jsonMedia(content *orderedmap.Map[string, *highv3.MediaType]) (string, *highbase.SchemaProxy, error) {
+	if content == nil || content.Len() == 0 {
+		return "", nil, nil
+	}
+	keys := make([]string, 0, content.Len())
```
daveshanley: `keys` is built on every call and only read in the error message. Allocate it inside the failure branch.

generator/sdk/prepare.go
```diff
@@ -0,0 +1,264 @@
...
+			operationID := operation.OperationId
+			if len(selected) > 0 {
+				if _, ok := selected[operationID]; !ok {
+					continue
+				}
+			}
+			if prior, ok := seen[operationID]; ok {
+				return nil, fmt.Errorf("sdk: duplicate operationId %q at %s and %s %s", operationID, prior, strings.ToUpper(method), path)
+			}
+			seen[operationID] = strings.ToUpper(method) + " " + path
```
daveshanley: `strings.ToUpper(method)` appears four times across 81-87 and runs twice on the happy path; hoist it once above the `selected` check. `found` is also a strict subset of `seen` (same keys, added at the same point), so the final loop can check `seen` and `found` goes away.

generator/sdk/golang/options.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2026 Princess B33f Heavy Industries / Dave Shanley
+// SPDX-License-Identifier: MIT
+
+package golang
+
+import "github.com/pb33f/libopenapi/generator/sdk"
+
+// Options configures Go SDK generation.
+type Options struct {
```
daveshanley, line 12: There is no way to reach the model generator from here. `renderModels` hard-codes its options, so a consumer cannot set `WithFormatMapping`, `WithEnumConstants`, `WithGenerateYAMLTags`, or anything else on the models this SDK ships with. Add `Models []modelgen.Option` and append the SDK's required options after the caller's so the SDK invariants win.

generator/sdk/golang/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2026 Princess B33f Heavy Industries / Dave Shanley
+// SPDX-License-Identifier: MIT
+
+// Package golang emits an idiomatic, resource-oriented Go SDK from a prepared
+// OpenAPI client contract.
+package golang
```
daveshanley: `generator/golang` and `generator/sdk/golang` are both `package golang`. The emitter already has to alias its own sibling as `modelgen`, and every consumer that wants models plus a client imports two packages with the same name. I would rather this was `gosdk`, or lived under `generator/golang/sdk`, than make every import site pick aliases forever.

generator/sdk/model.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// PrepareOptions selects operations and controls their public names. An empty
+// Operations slice selects every operation with an operationId.
+type PrepareOptions struct {
+	Operations []string
+	Names      map[string]OperationName
+}
+
+// Diagnostic reports a source operation that could not be selected or prepared.
```
daveshanley: The doc says this reports "a source operation that could not be selected or prepared". Nothing in `Prepare` ever emits one; every problem there is a hard error. The only producer is the emitter mapping model-generator diagnostics, with `OperationID` and `Method` always empty and the `Code` smashed into `Message` as a prefix. Either give it a `Code` field and document it as a model-generation diagnostic, or drop the operation fields. Right now the type describes something that does not happen.

generator/sdk/golang/view.go
```diff
@@ -0,0 +1,104 @@
...
+	Source string
+}
+
+type parameterView struct {
+	Name          string
+	FieldName     string
+	Type          string
+	In            string
+	Required      bool
+	Style         string
```
daveshanley: `Style` and `AllowReserved` are never read by a template. `prepareOperation` already rejects anything that is not the default style, so by the time a parameter reaches the view they carry no information. `operationView.Description` (line 34) and `workflowView.Description` (line 56) are dead too. A view should carry exactly what the templates print and nothing else.

generator/golang/to_go.go
```diff
@@ -435,14 +452,27 @@ func (g *Generator) formatType(format, fallback string) string {
...
-	if ir != nil && ir.Nullable && nullablePointer {
-		return true
+	if nullable || optional {
+		return 1
 	}
-	return !required && optionalPointers
+	return 0
+}
+
+func shouldPointer(typ string, ir *SchemaIR, required, optionalPointers, nullablePointer bool) bool {
```
daveshanley: `shouldPointer` has no production caller after this change; `goType` calls `pointerDepth` directly. It is alive only because `coverage_test.go:274` still calls it. Delete it and point that test at `pointerDepth`.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+var clientTemplateSource string
+
+//go:embed templates/resources.tmpl
+var resourcesTemplateSource string
+
+//go:embed templates/workflows.tmpl
+var workflowsTemplateSource string
+
+// Generate prepares an OpenAPI document and emits a Go SDK.
+func Generate(document *highv3.Document, options Options) (*sdk.Result, error) {
```
daveshanley, line 50: `Generate` prepares the contract internally, but `Options.Workflows` can only come from `sdk.PrepareArazzo(doc, contract)` against a contract the caller already built. So anyone with workflows has already called `Prepare`, and this entry point prepares again and discards their contract. Either take the Arazzo document and workflow ids here and prepare both, or document that `Generate` is the no-workflow path.

generator/sdk/prepare.go
```diff
@@ -0,0 +1,264 @@
...
+		appendOrReplace(parameter)
+	}
+	for _, parameter := range operationParameters {
+		appendOrReplace(parameter)
+	}
+	return result
+}
+
+func prepareParameter(parameter *highv3.Parameter) *Parameter {
+	required := parameter.Required != nil && *parameter.Required
```
daveshanley: Path parameters are only ever required, but this reads `required` straight off the spec and nothing defaults it for `in: path`. A spec that leaves `required` off (plenty do) generates `if input.ID != nil { ... replacePathParameter ... }`, so a nil `ID` sends `GET /things/%7Bid%7D` to the server. I generated it to be sure. Force `Required = true` when `In == "path"` here, or reject it at the emitter with the same error style as cookies and `allowReserved`.

generator/sdk/prepare.go
```diff
@@ -0,0 +1,264 @@
...
+				prepared.Schemes = append(prepared.Schemes, &RequiredSecurityScheme{Name: name, Scopes: append([]string(nil), scopes...)})
+			}
+		}
+		result = append(result, prepared)
+	}
+	return result
+}
+
+func methodSeed(operationID, resource string) string {
+	for _, suffix := range []string{resource, strings.TrimSuffix(resource, "s")} {
```
daveshanley: This collapses the list/singular pair, which is the most common REST shape there is. `getGoals` with resource `Goals` strips `Goals` to `get`. `getGoal` misses `Goals`, matches the trimmed `Goal`, and also lands on `get`. Both become `Goals.Get` and generation fails with "both map to". Running it on `test_specs/asana.yaml` with default names fails outright, so every real spec needs a hand-written `Names` map before the README's "generate a Go client" story works. Strip only the exact resource token and give the singular form its own suffix, rather than trimming an `s` and matching that too.

generator/sdk/workflow.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	return result, nil
+}
+
+func prepareWorkflow(workflow *higharazzo.Workflow, operations map[string]*Operation) (*Workflow, error) {
+	if strings.TrimSpace(workflow.WorkflowId) == "" {
+		return nil, errors.New("workflowId is required")
+	}
+	if workflow.Inputs != nil || workflow.Outputs != nil || len(workflow.DependsOn) > 0 || len(workflow.SuccessActions) > 0 || len(workflow.FailureActions) > 0 || len(workflow.Parameters) > 0 {
```
daveshanley: `inputs` is the JSON Schema that *defines* the `$inputs.*` values the payload bindings read. Rejecting it means the only Arazzo documents this compiler accepts are ones that reference inputs they never declared, which is exactly what our own fixture does (`generator_test.go:444-462`, three `$inputs.*` references and no `inputs` block). It is also why `resolveWorkflowFieldTypes` exists at all: with no input schema the emitter has to reverse-engineer input types out of its own generated Go. Honour `workflow.Inputs`, hand it to `generator/golang` for the input struct, and that whole parser round trip goes away.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+			return false
+		}
+	}
+	if nonNullType == "array" {
+		return schema.Items != nil && schema.Items.IsA() && parameterSchemaSupported(schema.Items.A)
+	}
+	return nonNullType != "" || len(schema.Enum) > 0
+}
+
+func statusIsSuccess(status string) bool {
```
daveshanley, line 961: `statusIsSuccess` takes any three-character key starting with `2`, so a spec written with a lowercase `2xx` (there are a lot of those in the wild) generates `statusMatches(response.StatusCode, "2xx")`. The generated `statusMatches` only knows uppercase `XX`, so it falls through to `fmt.Sprintf("%03d") == "2xx"`, which never matches. Every response, including a 200, becomes an `APIError`. Generated it to confirm. Uppercase the key in `prepareResponses`, or reject non-canonical keys with a proper error. Silent is the one thing this cannot be.

generator/sdk/model.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Contract is the effective client contract prepared from one OpenAPI document.
+// It retains schema proxies so language emitters can reuse libopenapi's model
+// generators without introducing a second schema representation.
+type Contract struct {
+	Title           string
+	Servers         []string
+	Operations      []*Operation
+	SecuritySchemes []*SecurityScheme
+	Schemas         *orderedmap.Map[string, *highbase.SchemaProxy]
```
daveshanley, line 22: This is the whole `components/schemas` map by pointer, and every `SchemaProxy` in it holds `idx *index.SpecIndex`, both YAML nodes and an atomic `rendered` cache. So a `Contract` that outlives its document pins the index, the rolodex and the full YAML tree, and every `proxy.Schema()` the emitter calls during generation leaves a rendered `Schema` cached on the *caller's* proxies. Generating an SDK measurably inflates the source document. I do not want a deep copy here. I want the lifetime written down: `Contract` is a borrowed view over the document, drop them together, and emitters should expect generation to warm the caller's caches. `Title` on line 16 is set by `Prepare` and read by nothing but tests.

generator/sdk/golang/generator.go
```diff
@@ -0,0 +1,770 @@
...
+	if components == nil {
+		components = orderedmap.New[string, *highbase.SchemaProxy]()
+	}
+	emitter := &emitter{
+		contract: contract, packageName: packageName, componentSchemas: components,
+		schemas: orderedmap.New[string, *highbase.SchemaProxy](), typeNames: make(map[string]string),
+		names: names, collectedComponents: make(map[string]struct{}),
+		visitedSchemas: make(map[*highbase.SchemaProxy]struct{}), workflows: workflows,
+	}
+	for name := range components.FromOldest() {
```
daveshanley, line 169: Every component in the document pays `PublicName` and reserves a name here, reachable or not. On a one-operation spec with two unused components: two names claimed, zero models emitted. The cost is small; the effect is not. An unreachable component can push a real inline type to `FooModel` just by existing. I know the eager claim is deliberate (`TestGenerateReservesComponentNamesBeforeCollectingInlineSchemas` wants component names to win), and lazy claiming inside `ensureComponent` would break that. The right shape is a reachability pre-pass: walk the selected operations first, claim names for the components you actually reached, then collect inline schemas.

generator/sdk/README.md
```diff
@@ -0,0 +1,97 @@
...
+The first emitter is `generator/sdk/golang`. It generates:
+
+- reachable models through `generator/golang`;
+- one shared HTTP client and security runtime;
+- resource-oriented methods such as `client.Accounts.List`;
+- typed success values, declared API errors, raw unknown errors, status and
+  response headers;
+- a narrow generated helper for supported finite Arazzo workflows.
+
+Playing Field is the first acceptance consumer. Product route names, roles,
```
daveshanley: This is a public library README. Consumer-specific route names and runtime behavior do not belong here, and they obscure the actual supported Arazzo subset. Describe the supported subset in specification terms and keep consumer notes in the consumer's repository.

daveshanley (commented):
Second pass, from a profiling run on `test_specs/asana.yaml` (151 operations, 135 components) rather than the PR's four-operation fixture. Seven new inline comments, four of them reproduced generation bugs: optional path parameters ship a literal `{id}` in the URL, `getGoals` and `getGoal` collapse onto one method so the large fixture fails to generate with default names, a declared Arazzo `inputs` block makes a workflow unsupported, and a lowercase `2xx` key generates a client where every response is an `APIError`. The other three concern retention (`Contract` pins the SpecIndex through `SchemaProxy`), names reserved for unreachable components, and consumer-specific product narrative in a public README.

Numbers at scale: one generation is 30ms, 21MB and 370k allocs. `format.Source` on the three template files is 61% of all allocation, which points at the structural lever missed in the first pass: everything lands in three monolithic files (`resources.gen.go` is 628KB for this fixture). One file per resource caps peak memory and lets formatting run in parallel.

One correction to the first pass: the `keys` slice in `jsonMedia` does not escape on Go 1.25, so that comment is a tidy-up rather than an allocation. The rest stands.

generator/sdk/golang/templates/resources.tmpl
daveshanley: Fixed on a0814798. Security requirement tables are emitted once at package scope, and generated-source coverage verifies the static declaration.

generator/sdk/golang/templates/resources.tmpl
daveshanley: Fixed on a0814798. Concrete and wildcard statuses now compile to integer and status-class predicates across resources and workflows; the runtime string matcher is gone.

generator/sdk/golang/templates/client.tmpl
daveshanley: Fixed on a0814798. Generated parameter code now uses typed string, integer, float, boolean, and array encoders with no reflection. Unsupported mapped scalar parameter formats fail generation explicitly.

generator/sdk/golang/templates/resources.tmpl
daveshanley: Fixed on a0814798. Response payload ownership is transferred without a second body copy, headers are cloned once, and bounded reads preallocate from a safe Content-Length cap.

generator/sdk/golang/templates/resources.tmpl
daveshanley: Fixed on a0814798. Query, body, and related imports and locals are emitted only for operations that use them.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. The SDK reuses exported RFC 6901 aware modelgen.RefName, accepts only exact local component schema refs, and rejects external refs explicitly until resolver support is complete.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Scalar type selection is delegated to the configured model generator. Mapped scalar parameters are rejected until a matching wire serializer is available, preventing type and wire divergence.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. The SDK emitter now uses the exported modelgen.NameRegistry with shared collision behavior and complete runtime identifier reservations.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. GeneratedType now exposes field and embedded-type metadata. Workflow binding resolution uses that metadata, including promoted allOf fields, with no generated-source AST round trip.

generator/sdk/workflow.go
daveshanley: Fixed on a0814798. Simple-condition splitting now lives in arazzo/expression and both the runtime and SDK compiler use the shared parser.

generator/sdk/workflow.go
daveshanley: Fixed on a0814798. Workflow payload and step parameter bindings use expression.Parse and reject unsupported nested expressions in the finite compiler.

generator/sdk/prepare.go
daveshanley: Fixed on a0814798. EffectiveStyle and EffectiveExplode now belong to highv3.Parameter and the SDK consumes those high-model methods.

generator/sdk/prepare.go
daveshanley: Fixed on a0814798. Document servers are resolved once and the prepared slice is shared by inheriting operations; only explicit path or operation servers are resolved separately.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Resource lookup uses stable indexes, and workflow lookup is built from pointers into the finalized operation slices, removing the linear scan and operation-view copies.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Embedded templates are parsed once at package initialization and quoting uses strconv.Quote.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. The coverage-only render injection and stage-failure seams were removed; tests exercise the public generation path.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Reachability follows only rendered shape keywords: properties, items, additionalProperties, allOf, oneOf, anyOf, and patternProperties.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. jsonMedia allocates and sorts media-type keys only on the unsupported-media error path.

generator/sdk/prepare.go
daveshanley: Fixed on a0814798. The method is normalized once and the redundant found map was removed in favor of the existing seen set.

generator/sdk/golang/options.go
daveshanley: Fixed on a0814798. Options.Models passes modelgen options through before SDK-required invariants, with mapped parameter formats rejected when no correct wire encoder exists.

generator/sdk/golang/doc.go
daveshanley: Fixed on a0814798. The SDK emitter package is named gosdk, removing the sibling package-name collision.

generator/sdk/model.go
daveshanley: Fixed on a0814798. sdk.Diagnostic now has stable Code, Path, and Message fields, mapped independently and covered by a regression test.

generator/sdk/golang/view.go
daveshanley: Fixed on a0814798. Dead style, reserved, and description fields were removed from the template views.

generator/golang/to_go.go
daveshanley: Fixed on a0814798. The production-dead shouldPointer helper was deleted and coverage targets pointerDepth directly.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Generate is explicitly the no-workflow entry point and rejects prebuilt workflows with guidance to Prepare, PrepareArazzo, and GenerateContract.

generator/sdk/prepare.go
daveshanley: Fixed on a0814798. Path parameters are always prepared as required, preventing unresolved placeholders from reaching generated requests.

generator/sdk/prepare.go
daveshanley: Fixed on a0814798. Default naming strips only the exact resource token, so singular and plural operation IDs remain distinct. Selected Asana operations generate and compile without overrides.

generator/sdk/workflow.go
daveshanley: Fixed on a0814798. Arazzo inputs are parsed as JSON Schema, handed to generator/golang for the public input model, and validated against direct step parameter and payload bindings. Undeclared, nested, and incompatible bindings fail generation.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. Wildcard response keys are normalized before classification and condition emission; lowercase 2xx is covered.

generator/sdk/model.go
daveshanley: Fixed on a0814798. Contract is documented as a borrowed view that shares schema proxies and may warm source caches; callers must keep and release it with the source document. The unused Title field was removed.

generator/sdk/golang/generator.go
daveshanley: Fixed on a0814798. A shape-only reachability prepass reserves names only for components reachable from selected operations, preserving inline names from unused components.

generator/sdk/README.md
daveshanley: Fixed on a0814798. The public README now describes only the general OpenAPI and Arazzo contract, supported subset, generated runtime, and library ownership boundaries.

daveshanley:
One review-summary suggestion remains intentionally deferred: emitting one file per resource.

I do not think the current profiling evidence justifies adding that artifact-layout change to this PR. Splitting the same generated source does not by itself reduce total `format.Source` work, and formatting resource files concurrently can increase peak memory. It also introduces a public filename-normalization and collision policy, deterministic file ordering rules, and a compatibility decision for `resources.gen.go` that are independent of the correctness and hot-path fixes here.

The current output is deterministic and the generated runtime cost is paid once at generation time. A separate change should define that file contract and include a large-spec benchmark showing an actual wall-time or peak-memory improvement before changing `Result.Files`.

Issue #628, opened 2026-09-18
`additionalOperations` is parsed into the low-level model but never reaches the high-level model


sheidkamp:
# Summary

An OpenAPI 3.2 document with `additionalOperations` loses those operations on a parse-and-render round trip. The low-level model parses them, but the high-level `PathItem.AdditionalOperations` stays nil, `GetOperations()` omits them, and they are absent from rendered output.

## Reproduction

```sh
mkdir /tmp/libopenapi-repro && cd /tmp/libopenapi-repro
go mod init repro
go get github.com/pb33f/libopenapi@v0.38.7

cat > main.go <<'EOF'
package main

import (
	"encoding/json"
	"fmt"

	"github.com/pb33f/libopenapi"
)

const spec = `{
  "openapi": "3.2.0",
  "info": {"title": "Repro", "version": "1.0.0"},
  "paths": {"/cache": {
    "get": {"summary": "Read the cache"},
    "additionalOperations": {"PURGE": {"summary": "Purge the cache"}}
  }}
}`

func cacheAdditionalOps(doc []byte) string {
	var d struct {
		Paths map[string]map[string]json.RawMessage `json:"paths"`
	}
	json.Unmarshal(doc, &d)
	if block, ok := d.Paths["/cache"]["additionalOperations"]; ok {
		return string(block)
	}
	return "(absent)"
}

func main() {
	doc, _ := libopenapi.NewDocument([]byte(spec))
	model, _ := doc.BuildV3Model()
	item, _ := model.Model.Paths.PathItems.Get("/cache")

	// The low-level model parsed the additional operations.
	fmt.Println("low-level model additionalOperations:")
	for pair := item.GoLow().AdditionalOperations.Value.First(); pair != nil; pair = pair.Next() {
		fmt.Printf("  %s: %q\n", pair.Key().Value, pair.Value().Value.Summary.Value)
	}

	// The high-level model never received them.
	fmt.Println("\nhigh-level model:")
	fmt.Print("  AdditionalOperations:")
	if item.AdditionalOperations == nil {
		fmt.Print(" nil")
	}
	for pair := item.AdditionalOperations.First(); pair != nil; pair = pair.Next() {
		fmt.Printf(" %s", pair.Key())
	}
	fmt.Println()
	fmt.Print("  Operations returned by GetOperations():")
	for pair := item.GetOperations().First(); pair != nil; pair = pair.Next() {
		fmt.Printf(" %s", pair.Key())
	}
[...]

PR #629, opened 2026-09-18
fix: AdditionalOperations handling

sheidkamp:
# Problem
https://github.com/pb33f/libopenapi/issues/628 - `additionalOperations` from the OpenApi 3.2 spec are dropped during a parse/render round-trip.

## Cause
`PathItem.Build` only set `Value` on `AdditionalOperations`, leaving `KeyNode` and `ValueNode` nil. `NodeReference.IsEmpty()` only looks at the nodes, so the high-level model treated the field as empty even when the map had operations in it. Any spec parsed from YAML silently lost its additional operations. They are missing from the high-level `PathItem`, from `GetOperations()`, and from rendered output.

## Fix
The fix captures the key and value nodes when the map is first created and sets them on the reference. With an `additionalOperations` container, that's the container's own key and value. With bare method keys there is no container, so it's the first additional operation found.

## Testing
The two existing tests in `datamodel/high/v3` both parse the YAML through `Build()` but were hiding the bug.

- `TestPathItem_WithAdditionalOperations` replaced what `Build()` produced with a hand-built low-level reference that had the nodes set, under a comment saying `Build()` doesn't handle `additionalOperations`. It does, it just didn't set the nodes. The hand-built block was removed and the test now uses the reference `Build()` returns.
- `TestPathItem_AdditionalOperations` wrapped every assertion in an `IsEmpty()` check that, due to the bug, prevented the meaningful assertions from running. The check is removed and replaced by an assertion that it is NOT empty instead of skipping the AdditionalOperations validation.


daveshanley:
LGTM!

